PR #4, opened 2015-04-23
fence_scsi: status should return off if any device is off

johnruemker:
As-is, get_status will return on if any device is on, with the result being that the library will choose to not attempt to unfence devices during an "on" action even if some devices need it.  This can be seen by adding a new device to the list and running "on", which will simply display "Success: Already ON".  We should return off if any device is off.

### DIFF
--- a/fence/agents/scsi/fence_scsi.py
+++ b/fence/agents/scsi/fence_scsi.py
@@ -32,6 +32,9 @@ def get_status(conn, options):
 		else:
 			logging.debug("No registration for key "\
 				+ options["--key"] + " on device " + dev + "\n")
+			if options["--action"] == "on":
+				status = "off"
+				break
 	return status
 
 


marxsk:
@johnruemker : I don't agree with you in this case. The default behaviour is consistent with other fence agent (e.g. PDU) where one ON means that machine is still in cluster and can damage data. 

I agree that you use-case is different but is there any reason to map together SCSI devices instead of having separate mapping? There are people who uses fence_scsi with multipath and that's something what may break their configuration. Perhaps we can make an option that will switch these behaviours.

johnruemker:
Good point, I hadn't thought enough about the other cases.  I had considered just making the recommendation about using separate devices, and the major concern there is that it will be difficult to reach all users with such a recommendation, so we'll still have many that will use a single device with one list.  And the problem should still occur even when using clvmd auto-detection, and so it'd be a burden to convince everyone to switch to specifying their devices just to get around this. 

This needs more thought about the best approach.

johnruemker:
How about this?  I can't think of any reasons why this approach won't work.  Just limiting the previously submitted behavior to only the "on" action, otherwise we keep the original behavior of reporting on if any device is on.

marxsk:
I agree with idea of this solution but it has to be changed a bit, possibly with another variable.

Let assume that we have old device OLD and new device NEW.
If they are ordered OLD, NEW -> it works as expected
If they are ordered NEW, OLD -> NEW will set status to "off" but the OLD will replace it by "on"

johnruemker:
Done.  Thanks again for your review and guidance.

PR #5, opened 2015-05-15
Fix bad whitespace in lpar, mpath and zvmip agents

ChristophBerg:
Without this, python is unhappy about the files


marxsk:
Thanks, fixed

PR #6, opened 2015-05-19
fence_rhevm fixes

rtheys:
Hi,

Please consider pulling these fixes for the fence_rhevm agent.

Commit 03eee86 adds the Filter header to the curl command. Without this the user used to connect to oVirt/RHEV-M needs more privileges than the UserRole role. With the filter header set, the UserRole permissions are sufficient to make the fence agent work.

Commit d19bb9b fills in the status of the outlets to prevent a crash on the monitor action. This error only occurs when the list of vm's returned is non-empty (when a user with UserRole role is used).

Regards,

Rik


marxsk:
Thanks, fixed

PR #9, opened 2015-07-11
fence_virsh: escape "EXPECT" string.

disco-stu:
Escape EXPECT string to prevent nroff error "warning: can't find special
character `EXPECT\'`"

Signed-off-by: Adrian Vondendriesch <adrian.vondendriesch@credativ.de>


marxsk:
Thanks for patch

PR #11, opened 2015-07-28
fence_rcd_serial: update description, makefile & add circuit diagram

sammcj:
As per https://bugzilla.redhat.com/show_bug.cgi?id=1240868#c15

Also added serial diagram as SVG into the source which renders to:

rcd_serial_diagram


marxsk:
ok, thanks

PR #15, opened 2015-08-25
fence_ilo: Default to tls1.0

johnruemker:
The latest iLO firwares (2.27 and later) disable SSLv3 by default and
fail at auto-negotiating TLS versions.  Forcing TLSv1.0 results in a
working connection by default.

### DIFF
--- a/fence/agents/ilo/fence_ilo.py
+++ b/fence/agents/ilo/fence_ilo.py
@@ -74,6 +74,7 @@ def main():
 	all_opt["login_timeout"]["default"] = "10"
 	all_opt["retry_on"]["default"] = "3"
 	all_opt["ssl"]["default"] = "1"
+	all_opt["tls1.0"]["default"] = "1"
 
 	options = check_input(device_opt, process_input(device_opt))
 


johnruemker:
My approach is flawed, as it doesn't leave the option for users to disable tls1.0 for fence_ilo[2] if they so choose (such as if the firmware begins to support later versions, or negotiation starts working).  Doing it my way require switching --tls1.0 to an argument-based attribute, which would then potentially break for some users using the old syntax.  

My temptation is to suggest adding yet another tls option, --tls1.0-fallback or similar, that fence_ilo would enable and the lib would use as reason to first try standard negotiation, then switch to using the TLSv1.0 option string if that failed.  However this would create a third tls-based option, further confusing users about which one to use, and it would also unnecessarily lengthen the time it takes to fence with fence_ilo because we'd have to wait for the auto-negotiation connection to fail first.

So, I don't have a great answer.  I'm going to close this pull request for now and follow up.

marxsk:
Hi,

your original suggestion is good but changing default behaviour should not occur on 4.0.(X+1). Fallback to other option should be fine. It will not require any changes from the user perspective. IWe will just have to check if it is not too slow - imho it will be fine.

PR #22, opened 2015-10-11
fence_vbox: Add a fencing agent for VirtualBox

krig:
Adds a fencing agent for virtual machines managed by VirtualBox.

Based on fence_virsh, tested with VirtualBox version 5.0.4 on
openSUSE 13.2.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -311,6 +311,7 @@ AC_CONFIG_FILES([Makefile
 		 fence/agents/sbd/Makefile
 		 fence/agents/sanbox2/Makefile
 		 fence/agents/scsi/Makefile
+		 fence/agents/vbox/Makefile
 		 fence/agents/virsh/Makefile
 		 fence/agents/vmware/Makefile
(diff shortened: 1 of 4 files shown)


digimer:
Glad to see this! The more environments we support, the better. :D

marxsk:
Hi krig,

thanks for fence agent. I have just two comments that may be worth fixing:
* line73: time.sleep(...) -> this should be done in fencing library. if it does not work it is my bug, just let me know. 
* You have used 'missing_as_opt' option which really makes sense (and I should port it to fence_virsh too) but this option has to be turned on by user (--missing-as-off). This is not a case for your code, simple 'if "--missing-as-off" in options' should solve this issue. 

I will wait for next upstream release, so it can be included.

krig:
@marxsk Good catch, thank you! Pushed fixes to submit request.

marxsk:
Great thanks, last very minor issue :) please merge it to one commit. thanks,

krig:
@marxsk done :)

PR #23, opened 2015-10-12
Set mark host down flag if possible

sbauza:
DO NOT MERGE YET: I have a question below, please answer it before we merge that proposal, or it would prevent an host to bring back automatically when the service is up.

Since Nova API version 2.11, there is now a new novaclient tool
that allows to bypass the servicegroup check and not wait for it
to catch the right host liveness. Using it as a first try and if
failing, fallbacking to the existing polling.

### DIFF
--- a/fence/agents/compute/fence_compute.py
+++ b/fence/agents/compute/fence_compute.py
@@ -92,15 +92,28 @@ def set_power_status(_, options):
 			override_status = "on"
 		return
 
-	# need to wait for nova to update its internal status or we
-	# cannot call host-evacuate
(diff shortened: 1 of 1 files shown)


sbauza:
You'll see a TODO asking for implementing the force_down=False failback when the host is back up.

I'm wondering where to land the opposite command, I'm guessing below https://github.com/ClusterLabs/fence-agents/blob/master/fence/agents/compute/fence_compute.py#L89 ?

sbauza:
Will provide the updated changeset in a separate pull request since I can't amend it directly.

PR #25, opened 2015-10-13
Allow to select which instances to evacuate

sbauza:
In order to select which instances could be evacuated automatically,
the instance has to have either:
 - a flavor where one extra spec key is "evacuate" with "True" as value
 - an image where a metadata key is "evacuate" with "True" as value

In case one instance is neither having a flavor or an image like that,
the instance won't be evacuated.


sbauza:
Updated PR with amended behaviour (by default, a host is fully evacuated unless there are instances with flavor/image asking for evacuation)

PR #28, opened 2015-11-21
fence_ipmilan: add a new method diag for the reboot action

actatux:
In some context, it can be useful to trigger kdump on a host while fencing it.
This patch allows to combine the reboot action with the method 'diag', based
upon the existant method 'cycle'.

### DIFF
--- a/fence/agents/ipmilan/fence_ipmilan.py
+++ b/fence/agents/ipmilan/fence_ipmilan.py
@@ -27,6 +27,10 @@ def reboot_cycle(_, options):
 	output = run_command(options, create_command(options, "cycle"))
 	return bool(re.search('chassis power control: cycle', str(output).lower()))
 
+def reboot_diag(_, options):
+	output = run_command(options, create_command(options, "diag"))
+	return bool(re.search('chassis power control: diag', str(output).lower()))
+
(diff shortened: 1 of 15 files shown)


marxsk:
Hi,

yours patch impact more fence agents than is required. The other problem is that --action=diag was used before so we have to respect that. I have created a new patch: https://github.com/ClusterLabs/fence-agents/commit/7e65180162a54ff8c6d4c883be73376a3385dfe7 that should solve this problem

actatux:
While I do agree it impact more fence agents than required, I found it more coherent regarding the cycle method that was already implemented.
My first approach was the same than commit 7e65180. The second one was to keep action=reboot and add a diag options (--diag --action=reboot). Finally, I opted for this version because it is the one that make more sense (to me, obviously).

What action do we want to perform? reboot the fenced node
What method do we want to use? reboot (default), cycle, diag, or whatever in the future

With good documentation, any user should be able to configure the fence agent to diag.

Anyway, we can now diag with the new patch, thanks.

marxsk:
@actatux 
I agree, having 'diag' in cycle will be fine but losing it from action was a regression that has to be fixed.

PR #33, opened 2015-12-03
Instance HA fixes/improvements

beekhof:


beekhof:
Tests fail due to intended metadata change

marxsk:
It should be enough to do 'make xml-upload.fence_compute' in fence agent directory.

PR #36, opened 2015-12-18
Fix list of servers per host

sbauza:
The previous query was incorrect, we need to use the 'host' keyname for getting
the list of instances per host as it's specified in
http://developer.openstack.org/api-ref-compute-v2.1.html#listServers


vuntz:
FWIW, this was fixed in https://github.com/ClusterLabs/fence-agents/commit/50335a68977670eef586eb8ab7a5be940db52519 already.

beekhof:
ok, closing

PR #42, opened 2016-02-03
fence_compute: Create nova client with API 2.11

vuntz:
The force_down API requires to use version 2.11 or later.

### DIFF
--- a/fence/agents/compute/fence_compute.py
+++ b/fence/agents/compute/fence_compute.py
@@ -348,7 +348,7 @@ def main():
 			sys.exit(0)
 
 	# The first argument is the Nova client version
-	nova = nova_client.Client('2',
+	nova = nova_client.Client('2.11',
 		options["--username"],
 		options["--password"],
 		options["--tenant-name"],
--- a/fence/agents/compute/fence_compute.py
+++ b/fence/agents/compute/fence_compute.py
@@ -324,7 +324,12 @@ def main():
 	run_delay(options)
 
 	try:
-		from novaclient import client as nova_client
+		from distutils.version import LooseVersion
+	except ImportError:
+		fail_usage("distutils not found or not accessible")
+
+	try:
+		import novaclient
 	except ImportError:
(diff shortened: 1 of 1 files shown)


beekhof:
Is there a way we can have this fall back to plain '2' if 2.11 is not available?
(Safe since we drop back to the wait loop if the force_down call fails)

beekhof:
novaclient.exceptions.UnsupportedVersion: Invalid client version '2.11'. must be one of: 3, 2, 1.1

vuntz:
> novaclient.exceptions.UnsupportedVersion: Invalid client version '2.11'. must be one of: 3, 2, 1.1

Hrm, so you use an old version of novaclient, I guess. I added a commit that should make it work.

aspiers:
@vuntz CI needs fixing somehow.

oalbrigt:
The issue has been fixed in master. Can you rebase and patch the updated master branch?

vuntz:
> The issue has been fixed in master. Can you rebase and patch the updated master branch?

Thanks, done.

fence/agents/compute/fence_compute.py
```diff
@@ -347,8 +352,12 @@ def main():
 		elif options["--action"] in ["monitor", "status"]:
 			sys.exit(0)
 
-	# The first argument is the Nova client version
-	nova = nova_client.Client('2',
+	if LooseVersion(novaclient.__version__) <= LooseVersion('2.26.0') :
```
beekhof: Is that going to do anything useful if LooseVersion couldn't be imported?
Surely the agent would just throw an error and abort

vuntz: See the `fail_usage` call earlier if it cannot be imported...

vuntz:
I think this patch is obsoleted by aa8b29615274feb0e1904b9a887a11ac89d95d24. I don't get why my patch wasn't merged months ago, though...

PR #43, opened 2016-02-03
fence_compute: Rework how the domain option works

vuntz:
Right now, we always append the domain to the host that we get and hope
for the best. This is not optimal and breaks in the cases where nova is
listing hypervisors with FQDNs but services with short hostnames.

Since the domain option is used when listing the hypervisors to also
list hypervisors with short hostnames, we know that the NovaEvacuate
agent will be able to match both FQDNs and short hostnames.

So what really needs to work fine is when we do actions on the nova
services. We change the logic to actually try with the hostname we're
given (which can be a FQDN or a short hostname); if there's no services
on that hostname, we try to use the domain option to build the
alternative hostname (FQDN if what we had was a short hostname; short
hostname if what we had was FQDN), and work with this instead.

This should allow all mixes of FQDN / short hostnames between nova and
pacemaker.

### DIFF
--- a/fence/agents/compute/fence_compute.py
+++ b/fence/agents/compute/fence_compute.py
@@ -22,6 +22,18 @@ nova = None
 EVACUABLE_TAG = "evacuable"
 TRUE_TAGS = ['true']
 
+def _get_alternative_host(host, options):
+	alt_host = None
+
+	# build alternative hostname, with the domain option: see if we got passed
+	# a short hostname or a FQDN, and give back the other one
+	if options["--domain"] != "":
+		alt_host = host.replace("." + options["--domain"], "")
+		if alt_host == host:
+			alt_host = host + "." + options["--domain"]
+
+	return alt_host
+
 def get_power_status(_, options):
 	global override_status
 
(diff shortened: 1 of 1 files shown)


vuntz:
I think https://github.com/ClusterLabs/fence-agents/pull/46 is actually the right way to go.

I'm leaving this open, though, in case there's a disagreement in https://github.com/ClusterLabs/fence-agents/pull/46 to deprecate the domain option.

beekhof:
I've significantly reworked the domain handling since, let me know if you still think this is necessary

vuntz:
> I've significantly reworked the domain handling since, let me know if you still think this is necessary

I need to look at the changes, but I still believe something like the first two commits of https://github.com/ClusterLabs/fence-agents/pull/46 are needed, since it's wrong to mix hypervisors and nova service hosts...

vuntz:
@beekhof so looking at https://github.com/ClusterLabs/fence-agents/commit/33e63ce9959199bdfdf8167d2f64a85776d30141: it's wrong again (mixing hypervisors, which are always FQDN, and nova service hosts, which are not always FQDN), and I think it's also buggy... Will submit a patch.

vuntz:
@beekhof see https://github.com/ClusterLabs/fence-agents/pull/80 for my patch

PR #44, opened 2016-02-03
fence_compute: Fix disabling force_down on node when action is on

vuntz:
When the action is on, the goal is to disable force_down for the
nova-compute service on the node.

However, we were only doing that if the nova-compute service was up;
which is impossible if it's forced to be down... So just always disable
force_down, and then, if it's up, do more things.


vuntz:
Hrm, it's probably not correct as that would re-enable the nova-compute too early. We probably need to do that from the NovaCompute OCF agent, and add a condition here (like evacuate attribute unset or set to no).

beekhof:
seems reasonable. @aspiers ?

vuntz:
Rebased, cc'ing @krcmarik since he did 5964f95ab698a1f5cf50af3654350eebb961eb61 not too long, but I feel it's not fully correct.

aspiers:
@vuntz I don't follow the logic here, but this code has changed loads since then anyway, so I'm not sure if your concern is already addressed. Please can you check?

PR #45, opened 2016-02-03
fence_compute: Evacuate instances on all tenants

vuntz:
We don't want to evacuate instances in just one tenant, so when we list
the instances, we need to look at all tenants, not just the current one.


beekhof:
Found this one independently, already upstream

PR #46, opened 2016-02-04
fence_compute: Do not list hypervisors but nova-compute services and deprecate domain option

vuntz:
There's no reason to list hypervisors, and this bit was actually, as far as I can tell, the only reason to have a domain option.

### DIFF
--- a/fence/agents/compute/fence_compute.py
+++ b/fence/agents/compute/fence_compute.py
@@ -34,7 +34,7 @@ def get_power_status(_, options):
 
 	if nova:
 		try:
-			services = nova.services.list(host=options["--plug"])
+			services = nova.services.list(host=options["--plug"], binary="nova-compute")
 			for service in services:
 				logging.debug("Status of %s is %s" % (service.binary, service.state))
 				if service.binary == "nova-compute":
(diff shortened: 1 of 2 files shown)


beekhof:
Deprecating the domain option isn't viable without adding some equivalent logic.
Without it, the `nova.services.list(host=options["--plug"], ...)` call returns nothing.

    >>> services = nova.services.list(binary="nova-compute")
    >>> for service in services: print service.host
    ... 
    overcloud-compute-3.localdomain
    overcloud-compute-0.localdomain
    overcloud-compute-1.localdomain
    overcloud-compute-2.localdomain
    >>> services = nova.services.list(host="overcloud-compute-0", binary="nova-compute")
    >>> for service in services: print service.host
    ... 
    >>> services = nova.services.list(host="overcloud-compute-0.localdomain", binary="nova-compute")
    >>> for service in services: print service.host
    ... 
    overcloud-compute-0.localdomain

vuntz:
@beekhof but why aren't you passing the fqdn to fence_compute? That would be a bug in your setup, imho.

beekhof:
@vuntz because the cluster is not configured to use FQDN for node names which has been considered a normal practice dating back to the Heartbeat package.

PR #47, opened 2016-02-06
Print fence_zvm --help to stdout and exit with status 0

ChristophBerg:
This makes fence_zvm --help conform with the other fence agents
behavior.


marxsk:
Thanks for a patch.

PR #61, opened 2016-02-28
added parallelism to make in .travis.yml in order to improve speed an…

krichter722:
…d test parallel build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,6 @@ before_install:
 script:
   - ./autogen.sh
   - ./configure
-  - make
-  - make check
+  - make -j16
+  - make -j16 check
   - PYTHONPATH=fence/agents/lib python fence/agents/lib/tests/test_fencing.py


marxsk:
Hi,

thanks for a patch. Did this really help? In 'make check' there is one slow action (10s) and few fast (~0.x). -j works correctly only on single directory so it won't help us running slow actions in parallel. imho it is required to modify makefiles a bit more.

krichter722:
It helped me track down a failure of `make` which occurred locally (and which I'm investigating further) which didn't occur with `make -j16` on travis which is a good indication that it's a local issue only. Adding `-j` to `make` should always be Pareto-optimal (my experience differs, of course ;)) but since people will invoke `make -j16 check` (maybe even have `MAKEFLAGS=-j16` in their environment) it should be tested in CI.

marxsk:
Complex rewrite of makefiles were done by @oalbrigt so I'm closing this PR.

PR #63, opened 2016-03-13
fence_ipdu: fix hardcoded FENCEAGENTSLIBDIR.

disco-stu:
FENCEAGENTSLIBDIR was hardcoded to "/usr/share/fence".


marxsk:
Thanks for a patch

PR #71, opened 2016-05-05
Explicitly use version 3 of the oVirt API

jhernand:
Version on 4 of the oVirt engine will support two versions of the API:
version 3, which will be backwards compatible with older versions of the
engine, and version 4, which won't be compatible. The default used by the
server will be version 4, but clients can explicitly request a version
of the API using the "Version" header or the "/v3" URL prefix. The
header is the preferred mechanism because it is just ignored by older
versions of the server. This patch modifies the oVirt fence agent so
that the path and version of the API are configuratble, and so that
their default values work correctly with version 3 compatibility mode of
the API.

Signed-off-by: Juan Hernandez juan.hernandez@redhat.com

### DIFF
--- a/fence/agents/rhevm/fence_rhevm.py
+++ b/fence/agents/rhevm/fence_rhevm.py
@@ -81,14 +81,24 @@ def send_command(opt, command, method="GET"):
 		url = "https:"
 	else:
 		url = "http:"
+	if opt.has_key("--api-path"):
+		api_path = opt["--api-path"]
+	else:
+		api_path = "/ovirt-engine/api"
 
-	url += "//" + opt["--ip"] + ":" + str(opt["--ipport"]) + "/api/" + command
+	url += "//" + opt["--ip"] + ":" + str(opt["--ipport"]) + api_path + "/" + command
 
 	## send command through pycurl
 	conn = pycurl.Curl()
 	web_buffer = io.BytesIO()
 	conn.setopt(pycurl.URL, url)
-	conn.setopt(pycurl.HTTPHEADER, ["Content-type: application/xml", "Accept: application/xml", "Prefer: persistent-auth", "Filter: true"])
+	conn.setopt(pycurl.HTTPHEADER, [
+		"Version: 3",
+		"Content-type: application/xml",
+		"Accept: application/xml",
+		"Prefer: persistent-auth",
+		"Filter: true",
+	])
 
 	if "cookie" in opt:
 		conn.setopt(pycurl.COOKIE, opt["cookie"])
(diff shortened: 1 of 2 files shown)


jhernand:
@marxsk, please review.

marxsk:
Hi,

there is a problem (found thanks to Travis otherwise I will miss it), you declare your new options as 'arg value' but fencing library works with them like they don't have arguments. Please add ':' to getopt values to mark that they require value (take a look at lib/fencing.py.py - line 176 for example).

m,

jhernand:
@marxsk I have added ":" to the options, and I also have verified that the agent works correctly with a live and real oVirt engine, version 4.

jhernand:
Fixed the help text of the new option, and added the default value. I believe tests will pass now.

jhernand:
@marxsk the Travis build is failing because it detects this difference between the generated and hand written XML:

```
30a31,35
> 	<parameter name="api_path" unique="0" required="0">
> 		<getopt mixed="--api-path=[path]" />
> 		<content type="string" default="/ovirt-engine/api"  />
> 		<shortdesc lang="en">The path of the API URL</shortdesc>
> 	</parameter>
75,79d79
< 	</parameter>
< 	<parameter name="api_path" unique="0" required="0">
< 		<getopt mixed="--api-path=[path]" />
< 		<content type="string" default="/ovirt-engine/api"  />
< 		<shortdesc lang="en">The path of the API URL</shortdesc>
```

But I don't see any real difference there. May it be a diff glitch? Any suggestion?

oalbrigt:
You can update the XML file by running the following command  in the fence/agents directory:
# make rhevm/fence_rhevm.xml-upload

and then you can verify that it's correct by doing:
# make rhevm/fence_rhevm.xml-check

jhernand:
Thanks @oalbrigt, that worked in my environment, lets see if it works in Travies.

By the way, I had to run the command inside the `rhevm` directory, and it was `make xml-upload.fence_rhevm`.

oalbrigt:
Can you rebase to fix the conflict?

jhernand:
Yes, done.

oalbrigt:
Thanks.

It seems like you need to go to fence/agents and do "make rhevm/fence_rhevm.xml-upload" to update the metadata for CI to succeed.

jhernand:
@oalbrigt , I did that, and I don't have conflicts in the master branch locally. But github still says that there are conflicting files. Really strange. Any suggestion?

jhernand:
@oalbrigt , looks like I managed to use an incorrect master branch, somehow. I am doing the rebase again, hopefully correctly this time.

jhernand:
@oalbrigt there was a long time since I tested this, please give me some time to check again that it works correctly before merging it.

oalbrigt:
I will. Thanks.

jhernand:
I just re-verified that this works correctly. However, I found an unrelated issue will doing it: the agent may pass `unicode` objects to pycurl, which rejects them. I have submitted another pull request to address that: ClusterLabs/fence-agents#106.

fence/agents/rhevm/fence_rhevm.py
```diff
@@ -136,9 +146,26 @@ def define_new_opts():
 		"required" : "0",
 		"shortdesc" : "Reuse cookies for authentication",
 		"order" : 1}
+	all_opt["api_path"] = {
+		"getopt" : "",
+		"longopt" : "api-path",
+		"help" : "--api-path                     The path part of the API URL",
```
oalbrigt: Can you add
"default" : "/ovirt-engine/api",
after the help-line?

You'll have to do the xml-upload again after doing so.

jhernand: @oalbrigt I added that line, but it doesn't have any effect on the generated XML. Is that OK?

oalbrigt: Oh. That's fine. I forgot that we ignore *_path in the tests to avoid issues when binaries are not located in the same location.

PR #73, opened 2016-05-10
fence_sbd: fix return value of fence functions

vvidic:
sbd command exits with 0 on success but fence functions
should return True in this case


marxsk:
Hi,

thanks for patch, I will merge it but can you please update a documentation in code, it still says that the return_code is integer instead of boolean.

m,

vvidic:
Code updated, please check again.

PR #75, opened 2016-05-25
this fix allows to use dedicated user for fencing, not only root@pam

xtavras:
More info to problem [1], this fixes the problem and allows to use dedicated user only with "VM.PowerMgmt" permission.

 https://forum.proxmox.com/threads/right-permissions-to-stop-vm-with-fence_pve.26618/#post-133621


marxsk:
ok, thanks

PR #80, opened 2016-06-29
fence_compute: Fix domain handling

vuntz:
https://github.com/ClusterLabs/fence-agents/commit/33e63ce9959199bdfdf8167d2f64a85776d30141 changed a bit how the domain is handled, but it comes with some issues:
- there are a number of bugs in the code
- it continues with the use of hypervisors (which are always FQDN) instead of moving everything to nova service hosts (which might not be FQDN) -- this matters because we actually only interact with nova service hosts (for the record, my first attempt to fix this was https://github.com/ClusterLabs/fence-agents/pull/46)

### DIFF
--- a/fence/agents/compute/fence_compute.py
+++ b/fence/agents/compute/fence_compute.py
@@ -34,7 +34,7 @@ def get_power_status(_, options):
 
 	if nova:
 		try:
-			services = nova.services.list(host=options["--plug"])
+			services = nova.services.list(host=options["--plug"], binary="nova-compute")
 			for service in services:
 				logging.debug("Status of %s is %s" % (service.binary, service.state))
 				if service.binary == "nova-compute":
--- a/fence/agents/compute/fence_compute.py
+++ b/fence/agents/compute/fence_compute.py
@@ -254,8 +254,8 @@ def fix_domain(options):
 		options["--domain"] = last_domain
 		return options["--domain"]
 
-	elif len(domains) == 1:
-		logging.error("Overriding supplied domain '%s' does not match the one calculated from: %s"
+	elif len(domains) == 1 and options["--domain"] != last_domain:
+		logging.error("Overriding supplied domain '%s' as it does not match the one calculated from: %s"
 			      % (options["--domain"], service.host))
 		options["--domain"] = last_domain
 		return options["--domain"]
--- a/fence/agents/compute/fence_compute.py
+++ b/fence/agents/compute/fence_compute.py
@@ -218,23 +218,23 @@ def fix_domain(options):
 	if nova:
 		# Find it in nova
 
-		hypervisors = nova.hypervisors.list()
-		for hypervisor in hypervisors:
-			shorthost = hypervisor.hypervisor_hostname.split('.')[0]
+		services = nova.services.list(binary="nova-compute")
+		for service in services:
+			shorthost = service.host.split('.')[0]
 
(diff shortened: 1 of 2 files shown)


vuntz:
cc @aspiers

aspiers:
LGTM on a skim read.

fence/agents/compute/fence_compute.py
```diff
@@ -34,7 +34,7 @@ def get_power_status(_, options):
 
 	if nova:
 		try:
-			services = nova.services.list(host=options["--plug"])
+			services = nova.services.list(host=options["--plug"], binary="nova-compute")
 			for service in services:
 				logging.debug("Status of %s is %s" % (service.binary, service.state))
 				if service.binary == "nova-compute":
```
beekhof: seems like we only need one of these checks for nova-compute

fence/agents/compute/fence_compute.py
```diff
@@ -218,72 +218,71 @@ def fix_domain(options):
 	if nova:
 		# Find it in nova
 
-		hypervisors = nova.hypervisors.list()
-		for hypervisor in hypervisors:
-			shorthost = hypervisor.hypervisor_hostname.split('.')[0]
+		services = nova.services.list(binary="nova-compute")
```
beekhof: reason?

vuntz: This is explained in the commit message: https://github.com/ClusterLabs/fence-agents/pull/80/commits/5cf13d5866d0541281c914ab4edf42cd09d5462b

fence/agents/compute/fence_compute.py
vuntz: done

vuntz:
@beekhof I addressed your comments and rebased.

vuntz:
@beekhof would it be possible to get this reviewed?

fence/agents/compute/fence_compute.py
```diff
@@ -229,24 +229,20 @@ def fix_domain(options):
...
-
 			if calculated == last_domain:
 				# Avoid complaining for each compute node with the same name
 				# One hopes they don't appear interleaved as A.com B.com A.com B.com
 				logging.debug("Calculated the same domain from: %s" % service.host)
+				continue
 
-			elif "--domain" in options and options["--domain"] == calculated:
-				# Supplied domain name is valid 
-				return
```
beekhof: I was thinking that we don't want to loop through all the compute nodes because there might be a LOT of them.
So I figured bailing early if we find at least one that had the supplied domain was sufficient

vuntz: But the warning about node with a different domain is important. And really, a loop over 1000 nodes will take less that 0.05s. That's much less than a call to the nova API.

vuntz:
Rebased one more time... It's slightly annoying to rebase, especially when changes like 19cb71c2e2685d019edb986b5aa86354107ab0ef or 07cedd7d996a50e71d85f09bf37aa68864c50db2 were for things that were already fixed in this patch.

beekhof:
Dammit, I tried to merge this but pushed to the wrong repo (mine, not ClusterLabs).
Will try again when I get back

aspiers:
@vuntz @beekhof These changes were all included in #142 so this can be closed now.

aspiers:
@vuntz @beekhof ping again, 5 months later :-)

oalbrigt:
Thanks for the update. Will do :)

PR #82, opened 2016-07-08
Fix for recursive make and some python 2.6 compat

lge:
Instead of, or in addition to pull request #81


oalbrigt:
Thanks.

I've added the Python 2.6 fixes and some of the "recursive make" changes to: https://github.com/ClusterLabs/fence-agents/pull/81

PR #84, opened 2016-08-10
build: fix possibly undefined macro AC_PYTHON_MODULE

jnpkrn:
As already mentioned 2 years back by Alan Evangelista[1], fence-agents
project indeed uses a strange build setup, which furthermore makes it
impossible to pass the configure stage successfully, at least with
some combinations of the "autotools" programs' versions (at least
with as old ones as present in RHEL 6).
Remedy is as simple as originally suggested, which also means m4
directory no longer needs to be reinsured in autogen.sh.

[1] https://www.redhat.com/archives/cluster-devel/2014-October/msg00161.html


jnpkrn:
@fabbione pointed out some possible deficiencies (there's issue with `.gitignore` for sure) so no rush with the merge, will broadcast when all seems fine.

jnpkrn:
Should be fine now.

jnpkrn:
@oalbrigt, @marxsk, ping

oalbrigt:
Works fine for me.

jnpkrn:
@marxsk, ping

autogen.sh
```diff
@@ -1,4 +1,3 @@
 #!/bin/sh
 # Run this to generate all the initial makefiles, etc.
-mkdir -p m4
-autoreconf -i -I make -v && echo Now run ./configure and make
+autoreconf -i -v && echo Now run ./configure and make
```
marxsk: was it intentionn to not include '-I make' ?

PR #92, opened 2016-11-07
fence_azure_arm: add fencing agent for Azure Resource Manager

MSSedusch:


jnpkrn:
@MSSedusch, it should be perfectly OK to stick with a single PR and just force push new iterations there, no need to consume PR namespace counter :-)

fence/agents/azure_arm/fence_azure_arm.py
```diff
@@ -0,0 +1,136 @@
...
+            return "on"
+
+    return "off"
+
+def set_power_status(conn, options):
+    logging.info("setting power status for VM " + options["--plug"] + " to " + options["--action"])
+
+    if compute_client:
+        rgName = options["--resourceGroup"]
+        vmName = options["--plug"]
```
MSSedusch: how can I move from plug to something like --vm? plug is filled automatically with the host that should be fenced

fence/agents/azure_arm/Makefile.am
```diff
@@ -0,0 +1,17 @@
+MAINTAINERCLEANFILES= Makefile.in
```
marxsk: The build system is now a bit different but that is something what we can easilly fix.

fence/agents/azure_arm/fence_azure_arm.py
```diff
@@ -0,0 +1,136 @@
...
+        "longopt" : "subscriptionId",
+        "help" : "-sid, --subscriptionId=[name]       Id of the Azure subscription",
+        "shortdesc" : "Id of the Azure subscription.",
+        "required" : "1",
+        "order" : 4
+    }
+
+# Main agent method
+def main():
+    global compute_client
```
marxsk: I would prefer to pass compute_client as an arg (instead of conn in other agents) but I can live with that :)

fence/agents/azure_arm/fence_azure_arm.py
```diff
@@ -0,0 +1,136 @@
...
+        vmName = options["--plug"]
+
+        if (options["--action"]=="off"):
+            logging.info("Deallocating " + vmName + "in resource group " + rgName)
+            compute_client.virtual_machines.deallocate(rgName, vmName)
+        elif (options["--action"]=="reboot"):
+            logging.info("Restarting " + vmName + "in resource group " + rgName)
+            compute_client.virtual_machines.restart(rgName, vmName)
+
+        sys.exit(0)
```
marxsk: Please, do not use sys.exit() in fence agent. Fencing library will test if fencing was succesfull or not.

fence/agents/azure_arm/fence_azure_arm.py
```diff
@@ -0,0 +1,136 @@
...
+    logging.info("setting power status for VM " + options["--plug"] + " to " + options["--action"])
+
+    if compute_client:
+        rgName = options["--resourceGroup"]
+        vmName = options["--plug"]
+
+        if (options["--action"]=="off"):
+            logging.info("Deallocating " + vmName + "in resource group " + rgName)
+            compute_client.virtual_machines.deallocate(rgName, vmName)
+        elif (options["--action"]=="reboot"):
```
marxsk: You will not enter 'reboot' part because reboot is translated to off, check if OFF, on, check if ON. If you want to have a reboot function that cannot be verified (like you have) you have to create a new function, pass it to fence_action(..) and add option --method (look at fence_ipmi)

MSSedusch:
The last commit should have solved review item 2 - 4. If you want me to fix item 1 as well, can you give me more details what you mean?
thanks

marxsk:
No, thanks. This is something what we can fix easilly.

marxsk:
Thanks a lot for new fence agent, it will be part of next upstream release.

fence/agents/azure_arm/fence_azure_arm.py
```diff
@@ -0,0 +1,131 @@
...
+            logging.info("Deallocating " + vmName + "in resource group " + rgName)
+            compute_client.virtual_machines.deallocate(rgName, vmName)
+        elif (options["--action"]=="on"):
+            logging.info("Starting " + vmName + "in resource group " + rgName)
+            compute_client.virtual_machines.start(rgName, vmName)
+
+
+def define_new_opts():
+    all_opt["resourceGroup"] = {
+        "getopt" : "rg:",
```
oalbrigt: These parameters need to be single character, as only single characters are supported for single dash parameters.

marxsk: @oalbrigt is right. Ideally remove this line completely.

PR #93, opened 2016-11-07
Add powerman agent

ofaaland:
Add a fence agent for use with powerman, https://github.com/chaos/powerman.


jnpkrn:
Do you really prefer to carry a dedicated and extra maintenance
requiring man page?  It's customary to have the man page automatically
generated based on the metadata, see `fence/agents/lib/fence2man.xsl`.

fence/agents/powerman/fence_powerman.py
```diff
@@ -0,0 +1,300 @@
...
+def get_list(conn, options):
+	logging.debug("get_list function:\noptions: %s", str(options))
+	pm = PowerMan(options['--ip'], options['--ipport'])
+
+	outlets = pm.list()
+	logging.debug("get_list outlets.keys: %s", str(outlets.keys()))
+	return outlets
+
+
+def define_new_opts():
```
marxsk: remove this function, if you don't need new options

fence/agents/powerman/fence_powerman.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+def main():
+	device_opt = [
+		'ipaddr',
+		'no_password',
+		'no_login',
+	]
+
+	raw_args=sys.argv
```
marxsk: This is something, what you don't want to have in the code. While I agree that it should be fine to see raw input and it can be added into fencing library, your solution did not count with using STDIN as an input method. And stdin is the standard way how to handle arguments in cluster environment.

fence/agents/powerman/fence_powerman.py
```diff
@@ -0,0 +1,300 @@
...
+	docs["shortdesc"] = "Fence Agent for Powerman"
+	docs["longdesc"] = "This is a Pacemaker Fence Agent for the \
+Powerman management utility that was designed for LLNL systems."
+	docs["vendorurl"] = "https://github.com/chaos/powerman"
+	show_docs(options, docs)
+
+	## Do the delay of the fence device 
+	run_delay(options)
+
+	if options["--action"] in ["off", "reboot"]:
```
marxsk: I don't understand a reason for this wait

fence/agents/powerman/fence_powerman.py
```diff
@@ -0,0 +1,300 @@
...
+	action = options["--action"]
+	if action == "on":
+		pm.on(options['--plug'])
+	elif action == "off":
+		pm.off(options['--plug'])
+
+	return
+
+
+def reboot(conn, options):
```
marxsk: You don't need this function if you don't use --method cycle (unconfirmable reboot). The 'standard' reboot function will work very similar to what you have with some extra features.

fence/agents/powerman/fence_powerman.py
```diff
@@ -0,0 +1,300 @@
...
+					return lst[0][:-1] # lst[0] would be 'off:'-- this removes the colon
+			# host isn't in the output
+			return -1
+
+
+def get_power_status(conn, options):
+	logging.debug("get_power_status function:\noptions: %s\n", str(options))
+	pm = PowerMan(options['--ip'], options['--ipport'])
+	# if Pacemaker is checking the status of the Powerman server...
+	if options['--action'] == 'monitor':
```
marxsk: Monitor action tests if fence device is running what you do test correctly. But 'response' off does not mean that fence device is off but it means that computer is off -> so monitor action will suceed. You have to return something else e.g. 'error' when PM is not running

fence/agents/powerman/fence_powerman.py
```diff
@@ -0,0 +1,300 @@
...
+		return False
+	time.sleep(2)
+	res = pm.on(options['--plug'])
+	if res < 0:
+		logging.error("reboot: power on failed.\n")
+		return False
+	return True
+
+
+def get_list(conn, options):
```
marxsk: This function (as referenced from fence_action) should return a lists of machines/plugs that can be turned ON/OFF. You don't have --plug option or any other way so I suspect that you can fence only one machine. Is it right?

fence/agents/powerman/fence_powerman.py
```diff
@@ -0,0 +1,300 @@
...
+			result, ret_code = self._run(cmd, True)
+		except OSError as e:
+			logging.error("PowerMan Error: The command '--on' failed: %s\n", e)
+			return -1
+		except ValueError as e:
+			logging.error("PowerMan Error: Popen: invalid arguments: %s\n", e)
+			return -1
+		logging.debug("pm.on result: %s ret_code: %s\n", result, ret_code)
+
+		## For some devices, the "on" or "off" command can fail silently,
```
marxsk: That is right point but we have support for this directly in fence library, so you don't have to worry about it.

fence/agents/powerman/fence_powerman.py
```diff
@@ -0,0 +1,300 @@
...
+	This class makes calls to a powerman deamon for a cluster of computers.
+	The make-up of such a call looks something like:
+		$ pm -h elssd1:10101 <option> <node>
+	where option is something like --off, --on, --cycle and where node is 
+	elssd8, or whatever values are setup in powerman.conf (***this is key, 
+	because otherwise this code will not work!)
+	"""
+	program_name = "powerman"
+
+	def __init__(self, server_name, port="10101"):
```
marxsk: magic number 10101 is not required here because you have defined default ipport so either it will be defined by user or default value will be used.

fence/agents/powerman/fence_powerman.py
```diff
@@ -0,0 +1,300 @@
...
+	"""Python wrapper for calling powerman commands
+
+	This class makes calls to a powerman deamon for a cluster of computers.
+	The make-up of such a call looks something like:
+		$ pm -h elssd1:10101 <option> <node>
+	where option is something like --off, --on, --cycle and where node is 
+	elssd8, or whatever values are setup in powerman.conf (***this is key, 
+	because otherwise this code will not work!)
+	"""
+	program_name = "powerman"
```
marxsk: Please, make the path to program configurable (e.g. like fence_ipmi & ipmitools). We attempt to not have constant across code.

fence/agents/powerman/fence_powerman.py
ofaaland: Do you mean I don't need this block of code:
`
                ## For some devices, the "on" or "off" command can fail silently,
                ## but for query powerman keeps trying until it gets a valid answer;
                ## so we must check after trying to power on/off a device.
                ## Powerman seems to report the old state if asked too quickly, wait a second.

                time.sleep(1)
                queryret = self.query(host)

                if (ret_code == 0 and queryret == "off") or (ret_code < 0 and queryret == "on"):
                        logging.warning("command '%s' returned (%s) but query reported (%s) afterwards\n", cmd, result, queryret)

                if queryret == "on":
                        return 0
                else:
                        return -1
`

and should just return ret_code from the "on" or "off" commands?

fence/agents/powerman/fence_powerman.py
ofaaland: powerman can fence a set of machines, e.g. `powerman --off wopr,pipboy[2000,3000]`.  It sounds like we've not implemented this properly.

fence/agents/powerman/fence_powerman.py
ofaaland: Me either.  This project was started by someone else.  I'll remove it and test.

fence/agents/powerman/fence_powerman.py
ofaaland: done.

fence/agents/powerman/fence_powerman.py
ofaaland: done

fence/agents/powerman/fence_powerman.py
ofaaland: retained it because I added --powerman-path

fence/agents/powerman/fence_powerman.py
ofaaland: I removed it here.  I will look at adding to library, with support for both stdin and argv, for debugging.

fence/agents/powerman/fence_powerman.py
ofaaland: get_list() does in fact return a dictionary of machines and their power state ("on" or "off") - all the machines that the powerman server knows about.
I thought that was what get_outlet_list function was supposed to do.  Did I get that wrong?

ofaaland:
I pushed an update that addresses all the feedback except for the question about get_list(), because I don't yet understand what is required.

ofaaland:
The test failed because fence_powerman was run and found that the powerman binary was missing.  Not surprising that you wouldn't have powerman installed.  How do you handle such dependencies?

fence/agents/powerman/fence_powerman.py
marxsk: yes, to be precise we don't care about return codes from set functions at all. Because quite often device really believes that everything is okay but it is not.

fence/agents/powerman/fence_powerman.py
marxsk: ok, so please add 'port' to device_opt array. This will enable --plug used to define which plug/machine/VM should we work with. (Yes, names are inconsistent but we can't fix them right now because of backward compatibility).

get_list - this is right,

fence/agents/powerman/fence_powerman.py
```diff
@@ -0,0 +1,286 @@
...
+def define_new_opts():
+	all_opt["powerman_path"] = {
+		"getopt" : ":",
+		"longopt" : "powerman-path",
+		"help" : "--powerman-path=[path]         Path to powerman binary",
+		"required" : "0",
+		"shortdesc" : "Path to powerman binary",
+		"default" : "/usr/bin/powerman",
+		"order": 200
+	}
```
marxsk: Please chnage default value to "@POWERMAN_PATH@" and add line to configure.in (line ~185 at root dir) and to make/fencebuild.mk 

This way the default can be distribution dependant.

fence/agents/powerman/fence_powerman.py
```diff
@@ -0,0 +1,286 @@
...
+	docs["shortdesc"] = "Fence Agent for Powerman"
+	docs["longdesc"] = "This is a Pacemaker Fence Agent for the \
+Powerman management utility that was designed for LLNL systems."
+	docs["vendorurl"] = "https://github.com/chaos/powerman"
+	show_docs(options, docs)
+
+	if not is_executable(options["--powerman-path"]):
+		fail_usage("Powerman not found or not executable at path " + options["--powerman-path"])
+
+	if options["--action"] in ["off"]:
```
marxsk: This can be defined by setting --power_wait option to different value. 

Look at https://fedorahosted.org/cluster/wiki/FenceTiming

ofaaland:
Pushed another update that addressed the comments from pass 2:
* default path for powerman set in configure
* remove redundant query within PowerMan.on() and off() used by set_power_fn
* use run_delay() before testing whether powerman binary exists and is executable

marxsk:
Thanks a lot for new fence agent, it will be part of next upstream release.

PR #95, opened 2016-11-21
fence_vmware_soap: Suppress InsecureRequestWarning when ssl_insecure is given

memotype:


marxsk:
Thanks, I have verified this on vCenter 5.5 with self-signed certificate.

PR #97, opened 2016-12-01
fencing: Specify command to be logged by run_command

n3world:
Currently run_command special cases the logging of commands with ipmitool and removes everything
between the flat -P and -p. Instead of special casing commands with ipmitool allow the caller to
provide the command to log so that the command can do any necessary scrubbing of the message to be
logged.

### DIFF
--- a/fence/agents/autodetect/fencing.py
+++ b/fence/agents/autodetect/fencing.py
@@ -901,17 +901,13 @@ def is_executable(path):
 			return True
 	return False
 
-def run_command(options, command, timeout=None, env=None):
+def run_command(options, command, timeout=None, env=None, log_command=None):
 	if timeout is None and "--power-timeout" in options:
 		timeout = options["--power-timeout"]
 	if timeout is not None:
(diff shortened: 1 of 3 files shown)


n3world:
It looks like the build failed because it couldn't find python library path, but I am not sure how my change would have caused that.

krig:
It is an unrelated failure, the fix in in pull request #96.

oalbrigt:
This seems to work as expected, and is to avoid logging the password. The reason for the "special casing" is that ipmitool uses -P for password and -p for port.

n3world:
The current code works ok for ipmitool but not for anything else. It also just replaces everything between -P and -p which could be other arguments and not just the password. If you look at fence_ipmilan you can see that the cipher argument can occur between -P and -p so that would be lost in the logging. This change was just to provide a more generic interface to scrubbing the logging of a command so that any agent can specify the log message for a command in case it also has a password or other sensitive information on the commandline.

PR #98, opened 2016-12-01
fencing: Add option to disable logging to stderr

n3world:
When using systemd and logging to stderr the journal captures both log messages and stderr together
causing duplicate messages.

Dec 01 16:38:48 d25-22-left fence_ipmilan[25856]: Executing: /usr/bin/ipmitool -I lan -H 172.20.71.75 -U admin -P [set] -p 623 -L ADMINISTRATOR chassis power status
Dec 01 16:38:48 d25-22-left fence_ipmilan[25856]: 0 Chassis Power is on
Dec 01 16:38:49 d25-22-left stonith-ng[23051]:  warning: fence_ipmilan[25856] stderr: [ Executing: /usr/bin/ipmitool -I lan -H 172.20.71.75 -U admin -P [set] -p 623 -L ADMINISTRATOR chassis power status ]
Dec 01 16:38:49 d25-22-left stonith-ng[23051]:  warning: fence_ipmilan[25856] stderr: [  ]
Dec 01 16:38:49 d25-22-left stonith-ng[23051]:  warning: fence_ipmilan[25856] stderr: [ 0 Chassis Power is on ]
Dec 01 16:38:49 d25-22-left stonith-ng[23051]:  warning: fence_ipmilan[25856] stderr: [   ]
Dec 01 16:38:49 d25-22-left stonith-ng[23051]:  warning: fence_ipmilan[25856] stderr: [  ]

Adding this option allows the user to disable the extra messages being logged to stderr.

### DIFF
--- a/fence/agents/autodetect/fencing.py
+++ b/fence/agents/autodetect/fencing.py
@@ -435,13 +435,20 @@ all_opt = {
 	"on_target": {
 		"getopt" : "",
 		"help" : "",
-		"order" : 1}
+		"order" : 1},
+	"quiet": {
+		"getopt" : "q",
+		"longopt": "quiet",
+		"help" : "-q, --quiet                    Disable logging to stderr. Does not affect --verbose or --debug logging to syslog.",
+		"required" : "0",
+		"order" : 50}
 }
 
 # options which are added automatically if 'key' is encountered ("default" is always added)
 DEPENDENCY_OPT = {
 		"default" : ["help", "debug", "verbose", "version", "action", "agent", \
-			"power_timeout", "shell_timeout", "login_timeout", "power_wait", "retry_on", "delay"],
+			"power_timeout", "shell_timeout", "login_timeout", "power_wait", "retry_on", \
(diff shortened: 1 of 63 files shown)


marxsk:
We have discussed this PR a bit. I have attempted to find a way how to do it right in pcs but it would have to be in pacemaker. So I agree with the proposed concept but I don't like the name of the option. What do you think about --quiet ?

n3world:
I am fine with --quiet. It would just have to be documented well that --quiet suppresses stderr output but does not affect syslog output so --debug or --verbose can be used with --quiet.

oalbrigt:
I think your description is fine enough for that, so if you update your code to --quiet it should be good.

n3world:
I ended up updating the help let me know if it is a bit too much.

oalbrigt:
Looks good to me. Can you change the commit-message to be the first line only (we only need a short description in the git log)?

oalbrigt:
Thanks.

PR #102, opened 2017-01-06
fence_powerman: fix typo in option powerman-path

ofaaland:
References to dictionaries device_opt[], options[] and all_opt[]
sometimes used key "powerman_path" and other times "powerman-path",
resulting in references to keys that did not exist.

This resulted in total failure of the fence agent.

This commit makes the key used consistent.

### DIFF
--- a/fence/agents/powerman/fence_powerman.py
+++ b/fence/agents/powerman/fence_powerman.py
@@ -185,7 +185,7 @@ def get_power_status(conn, options):
 
 def set_power_status(conn, options):
 	logging.debug("set_power_status function:\noptions: %s", str(options))
-	pm = PowerMan(options['--powerman_path'], options['--ip'], options['--ipport'])
+	pm = PowerMan(options['--powerman-path'], options['--ip'], options['--ipport'])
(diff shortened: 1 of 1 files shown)


marxsk:
Hi,

I agree that this has to be fixed but the situation is a bit more complex mostly because of backward compatibility issues. 

It is right to use --> options['--powerman-path']

But in all_opt / device_opt we use the variant with underscore - powerman_path. 

It looks strange but fence agents accept input also from STDIN where it was decided (too long ago) that we will use underscores so we follow this convention.  Of course, it leads to a situation when you have to enter different names on STDIN (or pcs) then on command-line :/ Situation is even a bit worse because names for stdin may differ from long getopt names massively (port=X vs --plug=X). We plan to fix this issue (probably still with underscore/dash issue), simplify documentation and still maintain backward compatibility.

ofaaland:
Thanks for the explanation.  Updated the commit to use underscore varient for all_opt device_opt as described.

oalbrigt:
Thanks.

PR #104, opened 2017-02-06
Add consistency to STDIN-getopt arguments

marxsk:
* Longopt options can be used on stdin either with dash or underscores
* Tests were not updated yet as there might be changes in XML metadata

### DIFF
--- a/fence/agents/vbox/fence_vbox.py
+++ b/fence/agents/vbox/fence_vbox.py
@@ -4,6 +4,7 @@
 #
 # VirtualBox 5.0.4 x64 on openSUSE 13.2
 #
+# @bug: I have to add additional log_expect() against previous version [mac test]
 
 import sys
 import re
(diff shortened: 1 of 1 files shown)


marxsk:
@oalbrigt please, take a look at code

fence/agents/vbox/fence_vbox.py
```diff
@@ -21,66 +22,82 @@
...
-    return options.get("--uuid") or options.get("--plug")
-
-_domain_re = re.compile(r'^\"(.*)\" \{(.*)\}$')
-
-
 def _invoke(conn, options, *cmd):
     prefix = options["--sudo-path"] + " " if "--use-sudo" in options else ""
-    conn.sendline(prefix + "VBoxManage " + " ".join(cmd))
+    conn.send_eol(prefix + options["--vboxmanage-path"] + " " + " ".join(cmd))
+    conn.log_expect(options["--command-prompt"], int(options["--shell-timeout"]))
```
oalbrigt: Should it check for it do the same log_expect twice?

marxsk: Aah, I just wanted to reuse branch. fence_vbox should not be part of this commit.

marxsk:
Applied in https://github.com/ClusterLabs/fence-agents/commit/de490e0590dabc8b30619092f9197ab0d4c47458

PR #106, opened 2017-02-08
Convert URLs passed to pycurl to ASCII

jhernand:
The 'fence_rhevm' agent uses pycurl to interact with the oVirt API, and
it builds the URLs using string concatenation. For example, to build the
URL used to start a virtual machine it does the following:

  url = "vms/" + options["id"] + "/" + action

This 'id' is extracted from the XML returned by the server, which may be
a plain 'str' or an 'unicode' object, depending on the version of pycurl
used. If it is an unicode object then resulting string will also be an
unicode object. When this unicode object is passed to pycurl it fails,
with the following traceback:

  Traceback (most recent call last):
    ...
    File "/usr/sbin/fence_rhevm", line 106, in send_command
      conn.setopt(pycurl.URL, url)
  TypeError: invalid arguments to setopt

The same happens with other agents. To avoid that this patch changes all
those agents so that they always encode the URL using ASCII before
passing it to pycurl.

Signed-off-by: Juan Hernandez <juan.hernandez@redhat.com>

### DIFF
--- a/fence/agents/cisco_ucs/fence_cisco_ucs.py
+++ b/fence/agents/cisco_ucs/fence_cisco_ucs.py
@@ -115,7 +115,7 @@ def send_command(opt, command, timeout):
 	## send command through pycurl
 	conn = pycurl.Curl()
 	web_buffer = io.BytesIO()
-	conn.setopt(pycurl.URL, url)
+	conn.setopt(pycurl.URL, url.encode("ascii"))
 	conn.setopt(pycurl.HTTPHEADER, ["Content-type: text/xml"])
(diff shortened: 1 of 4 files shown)


jhernand:
@oalbrigt please review.

oalbrigt:
Nice catch.

Can you solve the conflicts and update the pycurl.URL lines in the following agents as well?
cisco_ucs/fence_cisco_ucs.py
docker/fence_docker.py
pve/fence_pve.py
rhevm/fence_rhevm.py

jhernand:
Yes, done.

oalbrigt:
Thanks.

PR #109, opened 2017-02-14
Make pywsman dependency optional

krig:
If the pywsman python library is only imported when it is actually used,
it becomes possible to generate the man pages for the fence_amt_ws agent
even if it is not installed. That then means that there is no need to
require pywsman for the whole fence-agents package to be installable.

In our (SLE) case, pywsman is only available in the SDK module, so we
can't have a hard require on it in order to build or install the
fence-agents package for HA.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -163,7 +163,6 @@ AC_PYTHON_MODULE(suds, 1)
 AC_PYTHON_MODULE(pexpect, 1)
 AC_PYTHON_MODULE(pycurl, 1)
 AC_PYTHON_MODULE(requests, 1)
-AC_PYTHON_MODULE(pywsman, 1)
 
 ## path to 3rd-party binaries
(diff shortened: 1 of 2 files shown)


marxsk:
Sorry, I don't like this patch. But I agree that it should be possible to build it without pywsman. imho the right solution would be to require pywsman only when fence_amt_ws is going to be build. So, it should depends on ./configure --with-agents. 

The very same issue might be with python-suds and vmware_soap agent.

krig:
Yes, good point. Maybe we should start packaging each agent separately too.

Suds is also a problem since the project seems to be dead. Eventually it will need to be replaced...

oalbrigt:
This should fix the --with-agents issue: https://github.com/ClusterLabs/fence-agents/pull/110

krig:
Closing in favor of #110

PR #117, opened 2017-05-12
Workaround for invalid Pacemaker API call

winkelchri:
This fixes a invalid API call to Pacemaker's stonith-ng.

### DIFF
--- a/fence/agents/pve/fence_pve.py
+++ b/fence/agents/pve/fence_pve.py
@@ -179,6 +179,12 @@ machines acting as nodes in a virtualized cluster."
 	if options["auth"] is None:
 		fail(EC_LOGIN_DENIED)
 
+	# Workaround for unsupported API call on some Proxmox hosts
+	outlets = get_outlet_list(None, options)        # Unsupported API-Call will result in value: None
+	if outlets is None:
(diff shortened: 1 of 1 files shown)


fence/agents/pve/fence_pve.py
```diff
@@ -179,6 +179,12 @@ def main():
 	if options["auth"] is None:
 		fail(EC_LOGIN_DENIED)
 
+	# Workaround for unsupported API call on some Proxmox hosts
+	outlets = get_outlet_list(None, options)        # Unsupported API-Call will result in value: None
+	if outlets is None:
+		result = fence_action(None, options, set_power_status, get_power_status, None)
+		sys.exit(result)
+
```
oalbrigt: Looks good, but you can skip the if-section and change to outlets instead of get_outlet_list in the result-line below instead.

oalbrigt:
@winkelchri: ping... Can you update this PR as I suggested?

oalbrigt:
Oh. I'll look into that. Thanks.

fence/agents/pve/fence_pve.py
oalbrigt: You're right, so this seems to be the solution. Thanks.

PR #123, opened 2017-06-19
build: some cleanups to release.mk

jnpkrn:
Based upon https://github.com/ClusterLabs/libqb/pull/260

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,10 @@ make/stamp-h1
 /m4/*
 !/m4/ac_python_module.m4
 
+# make/release.mk related litter
+/.tarball-version
+/tag-*
+
 make/clusterautoconfig.h*
 missing
 *.pc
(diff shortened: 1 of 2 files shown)


oalbrigt:
Thanks.

Can you remove the \@echo-part regarding scp? We'll just do that if we decide to do it for the other projects.

jnpkrn:
On 22/06/17 02:51 -0700, Oyvind Albrigtsen wrote:
> Can you remove the \@echo-part regarding scp? We'll just do that if
> we decide to do it for the other projects.

Do you mean whole block starting with `Since` and ending with
`NOTE`?

There's actually a precedent of this "decision-pending" state in the
other project if you take a look at the referenced
https://github.com/ClusterLabs/libqb/pull/260

Or can I just comment that block out so it doesn't clutter the screen
for now?

Is currently `make -f make/publish.mk publish` being used at all?

-- 
Jan (Poki)

oalbrigt:
Yeah, you can just remove that block. We'll do the decision based on what is decided in e.g. libqb.

We still use the publish function.

jnpkrn:
OK, adjusted.

oalbrigt:
Thanks.

PR #129, opened 2017-07-16
Allow to override build date

bmwiedemann:
in order to allow for reproducible builds.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

### DIFF
--- a/scripts/fenceparse
+++ b/scripts/fenceparse
@@ -32,7 +32,7 @@ esac
 
 awk "{print}(\$1 ~ /#BEGIN_VERSION_GENERATION/){exit 0}" $realinfile
 echo ${start}RELEASE_VERSION=\"${release}\"${end}
-echo ${start}BUILD_DATE=\"\(built $(date)\)\"${end}
+echo ${start}BUILD_DATE=\"\(built $(date -u -d@${SOURCE_DATE_EPOCH:-$(date +%s)})\)\"${end}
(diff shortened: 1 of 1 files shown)


marxsk:
I'm not against the patch but I do not understand the benefits.

Will this work on RHEL 6.x (released 2010)? We still support python2.6 for that reason.

oalbrigt:
Looks good to me. The point is to be able to checksum a build and verify that it's the same package for extra security.

It should work fine in RHEL6.X, as it's not actually used except for as a comment in the files.

bmwiedemann:
Apart from security it has other benefits when you build a lot of packages. e.g. see
https://bugzilla.opensuse.org/show_bug.cgi?id=1047218#c0

marxsk:
@bmwiedemann ok, so your point is to detect changes in packages to prevent rebuilding. That's understandable.

I was talking to @oalbrigt and we agree that build time information is not useful for fence agents. It was included only for historical reasons (because it was always there). I will prepare a PR that will try to address your issue by removing that kind of information.

If you see any reason why such build time is useful, please let us know. I have never used it and it was never used in any customer reports that was handled to me.

bmwiedemann:
That is fine with me from the reproducible builds side.
@krig do you have an opinion from the maintainer side?

krig:
I'm positive to patches that improve the reproducibility of builds, so I'm all good with this. We don't have any need for the build information in the fence-agents themselves.

bmwiedemann:
closing now that #131 is merged

PR #130, opened 2017-07-19
Do not record BUILD_DATE

bmwiedemann:
to make build reproducible.
As the preferred alternative to PR #129


oalbrigt:
Looks good to me.

marxsk:
It looks good but I'm working on a bit bigger patch that remove a bit more. @oalbrigt feel free to decide if you wish that in a single or two patches. I will remove also copyrigth notice and simplify fenceparse (or remove it at best)

bmwiedemann:
IMHO in general it is preferable to have smaller patches - makes them easier to review and verify.
I have reviewed many hundred patches by now and the 1-4 liners are generally something you can do in under 5 minutes, but some of the bigger ones take weeks or months before you find the time.

bmwiedemann:
closing now that #131 is merged

PR #131, opened 2017-07-19
fencing: Remove unnecessary generated parts and support script

marxsk:
As @oalbrigt already commented, we should check info about copyright statement (and perhaps licensing as well).


make/fencebuild.mk
```diff
@@ -1,11 +1,9 @@
 define gen_agent_from_py
 	mkdir -p `dirname $@`
-	bash $(top_srcdir)/scripts/fenceparse \
-		$(top_srcdir)/make/copyright.cf REDHAT_COPYRIGHT \
-		$(VERSION) \
-		$(abs_srcdir) $@ | \
+	cat $@.py | \
```
oalbrigt: This needs to be updated to:
cat $(abs_srcdir)/$@.py | \

oalbrigt:
Thanks to @bmwiedemann who reported and provided the patch (https://github.com/ClusterLabs/fence-agents/pull/130) that led to this cleanup.

PR #142, opened 2017-09-13
OpenStack: Merge Vincent's changes and update agents for Pike

beekhof:
The split needs to be timed with an equivalent change to NovaEvacuate


fence/agents/compute/fence_evacuate.py
```diff
@@ -0,0 +1,372 @@
...
+
+sys.path.append("@FENCEAGENTSLIBDIR@")
+from fencing import *
+from fencing import fail_usage, is_executable, run_command, run_delay
+
+#BEGIN_VERSION_GENERATION
+RELEASE_VERSION="4.0.11"
+BUILD_DATE="(built Wed Nov 12 06:33:38 EST 2014)"
+REDHAT_COPYRIGHT="Copyright (C) Red Hat, Inc. 2004-2010 All rights reserved."
+#END_VERSION_GENERATION
```
oalbrigt: This section should be removed from both agents, as we've moved the essential part to lib/fencing.py to support reproducible builds.

https://github.com/ClusterLabs/fence-agents/pull/131/files#diff-94a84d829e5760b9a84e6d98bdba8fc4

oalbrigt:
CI fails due to changed metadata, so you should do a "make xml-upload" and commit the changes.

aspiers:
For reference, here is [the corresponding change to `NovaEvacuate`](https://review.openstack.org/#/c/503617/2).

PR #143, opened 2017-09-21
Do not provide tenant name in place for tenant ID.

jsuchome:
novaclient expects project ID as a positional argument, and
the fence_compute was providing tenant name instead.
Correct way of providing tenant name is with project_name key.


aspiers:
@jsuchome Thanks! But how did this ever work before? Has something changed in novaclient?

jsuchome:
I have no idea, it seems the change in nova-client is pretty old, as Colleen found out: https://github.com/openstack/python-novaclient/commit/9bbe5a87b7df62a7962debba5db7c96555da6761

beekhof:
I'll be merging https://github.com/beekhof/fence-agents/commit/88c1542454b1527fb24a7b3722fd1bf892b5ce07 soon which is more backwardly compatible

jsuchome:
Makes sense.

aspiers:
@beekhof to the rescue, hooray! :-)

PR #144, opened 2017-09-26
fence_heuristics_ping: add fencing agent for conditional fencing

wenningerk:
Not really a fencing agent by itself!
Put on the same fencing level with the actual fencing agent for
your node to make actual fencing depend on the result of (own)
connectivity determinded using ping heuristics.

resolves: rhbz#1476401

Tried to make that look as much as other fencing agents as possible
while making as much use of the fencing-library as possible.
This got a little hacky as the whole thing is quite a bit different from
the other fencing agents.
Changes to the fencing-library are minimal and shouldn't alter the
interface for existing fencing agents.

### DIFF
--- a/fence/agents/heuristics_ping/fence_heuristics_ping.py
+++ b/fence/agents/heuristics_ping/fence_heuristics_ping.py
@@ -18,22 +18,16 @@ from fencing import fail_usage, run_command, fence_action, all_opt
 from fencing import atexit_handler, check_input, process_input, show_docs
 from fencing import run_delay
 
-def get_power_fake(con, options):
-	# Hack to hand over an exit-code
-	return get_power_fake.result
-
 def ping_test(con, options):
 	# Send pings to the targets
 
 	if options["--action"] == "on":
 		# we want unfencing to always succeed
-		get_power_fake.result = "on"
 		return True
 
 	if not "--ping-targets" in options or options["--ping-targets"] == "":
 		# "off" was requested so fake "on" to provoke failure
 		logging.error("ping target required")
-		get_power_fake.result = "on"
 		return False
 
 	timeout = int(options["--ping-timeout"])
@@ -95,10 +89,6 @@ def ping_test(con, options):
 
 	if failcount > maxfail:
 		exitcode = False
-		# "off" was requested so fake "on" to provoke failure
-		get_power_fake.result = "on"
-	else:
-		get_power_fake.result = "off"
 
 	return exitcode
 
@@ -192,10 +182,10 @@ def main():
 	result = fence_action(\
 				None, \
 				options, \
-				ping_test, \
-				get_power_fake, \
 				None, \
-				ping_test)
+				None, \
+				reboot_cycle_fn = ping_test,
+				sync_set_power_fn = ping_test)
 
 	# execute the remaining delay
 	run_delay(options, result=result)
--- a/fence/agents/heuristics_ping/fence_heuristics_ping.py
+++ b/fence/agents/heuristics_ping/fence_heuristics_ping.py
@@ -1,0 +1,205 @@
+#!@PYTHON@ -tt
+
+# The Following Agent Has Been Tested On:
+#
+# RHEL 7.4
+#
+
+import io
+import re
+import subprocess
+import shlex
+import sys, stat
+import logging
+import os
+import atexit
+sys.path.append("@FENCEAGENTSLIBDIR@")
+from fencing import fail_usage, run_command, fence_action, all_opt
+from fencing import atexit_handler, check_input, process_input, show_docs
+from fencing import run_delay
+
+def get_power_fake(con, options):
+	# Hack to hand over an exit-code
+	return get_power_fake.result
+
+def ping_test(con, options):
+	# Send pings to the targets
+
+	if options["--action"] == "on":
+		# we want unfencing to always succeed
+		get_power_fake.result = "on"
+		return True
+
+	if not "--ping-targets" in options or options["--ping-targets"] == "":
+		# "off" was requested so fake "on" to provoke failure
+		logging.error("ping target required")
+		get_power_fake.result = "on"
+		return False
+
+	timeout = int(options["--ping-timeout"])
+	count = int(options["--ping-count"])
+	interval = int(options["--ping-interval"])
+	good_required = int(options["--ping-good-count"])
+	maxfail = int(options["--ping-maxfail"])
+	targets = options["--ping-targets"].split(",")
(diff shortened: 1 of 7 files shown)


fence/agents/heuristics_ping/fence_heuristics_ping.py
```diff
@@ -0,0 +1,195 @@
...
+		"shortdesc" : "The number of failed ping-targets to still account as overall success",
+		"default" : "0",
+		"order" : 1
+		}
+
+	all_opt["ping_targets"] = {
+		"getopt" : ":",
+		"longopt" : "ping-targets",
+		"required" : "1",
+		"help" : "--ping-targets=tgt1:tgt2       Colon separated list of ping-targets",
```
marxsk: we are using comma for setting multiple values at once (e.g. --plug 1,2,3)

fence/agents/heuristics_ping/fence_heuristics_ping.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def main():
+	device_opt = ["no_status", "no_password", "ping_count", "ping_goodcnt",
+		"ping_interval", "ping_timeout", "ping_maxfail", "ping_targets", "method"]
+	define_new_opts()
+	atexit.register(atexit_handler)
+
+	all_opt["method"]["default"] = "cycle"
+	all_opt["method"]["help"] = "-m, --method=[method]          Method to fence (onoff|cycle) (Default: cycle)"
```
marxsk: perhaps switch to (cycle|onoff) to emphasize cycle

fence/agents/lib/fencing.py.py
```diff
@@ -763,6 +763,9 @@ def set_multi_power_fn(connection, options, set_power_fn, get_power_fn, retry_at
 
 		for _ in range(int(options["--power-timeout"])):
 			if get_multi_power_fn(connection, options, get_power_fn) != options["--action"]:
+				if "no_status" in options["device_opt"]:
```
marxsk: If there is no 'status' there should be no 'return False', the status action should not be called at all.

wenningerk: Well, true in theory. But as implemented till now it is called and as the interface for setting a value doesn't foresee a return value this is an admittedly ugly way of still being able to convey a return value somehow for these cases where there is no real status-query available. I don't see an easy way to achieve that differently without breaking the interface.

fence/agents/heuristics_ping/fence_heuristics_ping.py
```diff
@@ -0,0 +1,195 @@
...
+	return exitcode
+
+
+def define_new_opts():
+	all_opt["ping_count"] = {
+		"getopt" : ":",
+		"longopt" : "ping-count",
+		"required" : "0",
+		"help" : "--ping-count=[number]          Number of ping-probes to send",
+		"shortdesc" : "The number ob ping-probes that is being sent per target",
```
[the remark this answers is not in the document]
wenningerk: careful reading ;-)

fence/agents/heuristics_ping/fence_heuristics_ping.py
```diff
@@ -0,0 +1,195 @@
...
+		"default" : "10",
+		"order" : 1
```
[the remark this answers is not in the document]
wenningerk: That was actually intentional. These are the parameters that define the statistics and none of them are more or less important in my mind. And users should be aware of the parameters and what goes into statistics calculation.

fence/agents/heuristics_ping/fence_heuristics_ping.py
```diff
@@ -0,0 +1,195 @@
...
+		}
+
+	all_opt["ping_goodcnt"] = {
```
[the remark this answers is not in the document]
wenningerk: No objection from my side. But if we go for readability I'm sparing an additional dash as well.

fence/agents/heuristics_ping/fence_heuristics_ping.py
```diff
@@ -0,0 +1,195 @@
...
+import sys, stat
+import logging
+import os
+import atexit
+sys.path.append("@FENCEAGENTSLIBDIR@")
+from fencing import fail_usage, run_command, fence_action, all_opt
+from fencing import atexit_handler, check_input, process_input, show_docs
+from fencing import run_delay
+
+PING_BINARY = "/bin/ping"
```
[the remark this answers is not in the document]
wenningerk: Doing it properly unfortunately requires handling of ping6 as well. Wanted to do that in a 2nd step but we can handle the path here for the first.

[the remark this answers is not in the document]
marxsk: if it will be fixed in next patch, I'm okay with it

fence/agents/heuristics_ping/fence_heuristics_ping.py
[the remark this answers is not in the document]
marxsk: ok, readability (understandability) counts

fence/agents/heuristics_ping/fence_heuristics_ping.py
wenningerk: this and going IPv6 using colons as separators isn't that wise either ...

fence/agents/lib/fencing.py.py
marxsk: I agree that it might require more work in fencing library but @oalbrigt should be able to fix that fast. We already have fence agent that have no_status (rcd_serial) and it looks like it works. But it was tested just by users.

wenningerk: I had of course checked rcd_serial! Reason why that is working is that it doesn't support setting an explicit value but just rebooting. Haven't found a fence-agent that supports explicit setting + no_status. With this kind of fake-status-call we can support that flavor without breaking the interface. 
Of course we could support a setter-function with return-value as enhancement of fence_action. (like result = fence_action(..., set_power_with_status=...)
Is that what you had in mind. I can try and see how clumsy that would finally look.

marxsk: That's a good point. We expect that verification cannot work with method=cycle but it was intended for reboot only originally.

wenningerk:
new commit should address all points on review except the no_status stuff

wenningerk:
last patch adds synchronous setting of power-state with return-value removing the necessity to use the existing interface in a hacky way.
Should just be an extension to the interface (fence_action) so there shouldn't be any compatibility issues with unchanged fence-agents.

PR #149, opened 2017-10-17
fencing: include timestamps when logging to STDERR and debug file

aspiers:
When logging to STDERR (i.e. when `--quiet` is not specified), or when logging to a debug file via `--debug-file`, include timestamps so that exact timings of fencing actions can be correlated with events logged elsewhere.


oalbrigt:
LGTM. Thanks.

PR #150, opened 2017-10-17
fencing: fix help for --quiet

aspiers:
`--quiet` erroneously referred to a `--debug` option which does not exist, and further implied that `--debug` affected logging to syslog, which is not even true of `--debug-file`, which logs to a file independently of logging to syslog.  So fix the text to clarify that `--quiet` affects neither logging to a debug file specified by `--debug-file`, nor logging to syslog.

### DIFF
--- a/fence/agents/autodetect/fencing.py
+++ b/fence/agents/autodetect/fencing.py
@@ -439,7 +439,7 @@ all_opt = {
 	"quiet": {
 		"getopt" : "q",
 		"longopt": "quiet",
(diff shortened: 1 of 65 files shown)


aspiers:
Seems that GitHub is currently broken - I just force-pushed [a new version of this](https://github.com/ClusterLabs/fence-agents/compare/master...aspiers:fix/quiet-help?expand=1) but this PR has not spotted it and updated accordingly :-/

oalbrigt:
Looks good to me. Thanks.

PR #151, opened 2017-10-18
Fix: fence_heuristics_ping: make fence-agent fail without params

wenningerk:
Without ping_targets given the agent will fail so unintentional usage as kind of a 'dummy'-fence-agent should happen less easily.


oalbrigt:
Looks good to me.

PR #152, opened 2017-10-18
Update drac5 to support drac8 ssh fence agent

Klaas-:
This adds a dell drac ssh fence agent, I tested it with drac8. I looked at the tests in tests/data/metadata/fence_ilo_ssh.xml - they should all apply to drac aswell but essentially testing the common parts once should be enough I'd guess :)

### DIFF
--- a/fence/agents/drac5/fence_drac5.py
+++ b/fence/agents/drac5/fence_drac5.py
@@ -8,6 +8,7 @@
 ## +-----------------+---------------------------+
 ##  DRAC 5             1.0  (Build 06.05.12)
 ##  DRAC 5             1.21 (Build 07.05.04)
+##  DRAC 8             2.43.43.43
 ##
 ## @note: drac_version was removed
 #####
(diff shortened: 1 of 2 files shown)


Klaas-:
seems like the tests are needed for automated testing to pass, I'll fix them tonight

Klaas-:
hmm I didn't notice the fence_drac5 is also ssh :) I'll integrate drac8 into that

oalbrigt:
Great. FYI you can run "make xml-upload" to update the test-metadata.

Klaas-:
its just a one line fix in the right file :D

Klaas-:
should I note somewhere that I had to significantly increase the shell/power timeout?

Klaas-:
@oalbrigt ping :)

marxsk:
@Klaas- yes, if is required for standard operation that you (or @oalbrigt ) can think about adding a new symlink with different default values, like we already do for fence_ipmi

PR #159, opened 2017-11-18
fence_gce: new agent

helen-fornazier:
First implementation of GCE agent

### DIFF
--- a/fence/agents/gce/fence_gce.py
+++ b/fence/agents/gce/fence_gce.py
@@ -1,0 +1,110 @@
+#!@PYTHON@ -tt
+
+import atexit
+import sys
+sys.path.append("@FENCEAGENTSLIBDIR@")
+from googleapiclient import discovery
+from oauth2client.client import GoogleCredentials
+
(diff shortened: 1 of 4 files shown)


fence/agents/gce/fence_gce.py
```diff
@@ -0,0 +1,110 @@
...
+		elif (options["--action"]=="on"):
+			conn.instances().start(project=options["--project"], zone=options["--zone"],
+							instance=options["--plug"]).execute()
+	# TODO: check which Exceptions it can throw
+	except :
+		fail_usage("Failed: Unable to connect to GCE. Check your configuration.")
+
+def define_new_opts():
+	all_opt["zone"] = {
+		"getopt" : "z:",
```
oalbrigt: Can you remove the short option here? We dont use those for new parameters as they often clash with parameters used in other agents.

helen-fornazier: Done

oalbrigt:
Thanks.

PR #163, opened 2017-12-11
Proposed fix to address the regex used to located longopt defined wit…

handruin:
…hin the individual fence agents code.  The previous regex does not account for PEP8 guidelines when defining dictionary key value separator spacing.  The defined stylization is to place the colon directly next to the quotation mark.  This regex parser misses this so the fix is to account for both space and no space for backward compatibility.  This parsing issue causes obscure issues when developing new fence agents using PEP8 indentation guidelines.


oalbrigt:
LGTM. Thanks.

PR #167, opened 2018-01-18
move fence_evacuate into its own subdirectory

aspiers:
The `configure.ac` script assumes that each `$agent` will live under `fence/agents/$agent/fence_$agent`, but `fence_evacuate` was violating that assumption which meant that `./configure` would fail when `AGENTS_LIST` was set to something other than `all`.

So move `fence_evacuate` into its own subdirectory to fix this.


oalbrigt:
Nice catch. Thanks.

PR #170, opened 2018-02-09
heuristics_ping: fix for python3 TypeError

vvidic:
Traceback (most recent call last):
  File "heuristics_ping/fence_heuristics_ping", line 198, in <module>
    main()
  File "heuristics_ping/fence_heuristics_ping", line 191, in main
    sync_set_power_fn = ping_test)
  File "lib/fencing.py", line 904, in fence_action
    if reboot_cycle_fn(connection, options):
  File "heuristics_ping/fence_heuristics_ping", line 75, in ping_test
    searchres = packet_count.search(line)
TypeError: cannot use a string pattern on a bytes-like object


oalbrigt:
Thanks.

PR #172, opened 2018-02-10
fence_powerman: use @PYTHON@ for interpreter

vvidic:
Using variable interpreter enables switch to python3.


oalbrigt:
Nice catch. Thanks.

PR #176, opened 2018-02-23
add new Azure fence agent

MSSedusch:
Add a new fence agent that is part of a new fencing mechanism for Azure virtual machines.
The fence operation is splitted into this fence agent (for off operation) and a resource agent (for on operation) to work around a limitation in pacemaker that makes the cluster wait for completion of the overall reboot operation.

Resource Agent PR: https://github.com/ClusterLabs/resource-agents/pull/1114


fence/agents/azure_arm/fence_azure_arm_network.py
```diff
@@ -0,0 +1,215 @@
...
+from fencing import *
+from fencing import fail, fail_usage, EC_TIMED_OUT, run_delay
+import time
+import azure_fence_lib
+
+#import ptvsd
+
+def get_azure_config(options):
+    config = azure_fence_lib.AzureConfiguration()
+
```
anandramakrishna: You should be able to get a bunch of this information from IMDS dynamically.  That way you don't need to have the user supply all this via configuration.  See https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service, and in particular, https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#instance-metadata-data-categories.

fence/agents/azure_arm/fence_azure_arm_network.py
```diff
@@ -0,0 +1,215 @@
...
+                azure_fence_lib.set_power_status_on(compute_client, network_client, rgName, vmName)
+        except Exception as e:
+            fail_usage("Failed: %s" % e)
+    else:
+        fail_usage("No Azure clients configured. Contact support")
+
+    logging.info("{set_power_status} done")
+
+def define_new_opts():
+    all_opt["resourceGroup"] = {
```
anandramakrishna: Same point as above about using instance metadata service to fetch this dynamically.

oalbrigt:
fence_azure_arm_network.py needs to be in it's own azure_arm_network-directory to be able to choose it with "./configure --with-agents=".

fence/agents/azure_arm/fence_azure_arm_network.py
MSSedusch: Great idea! I made resourceGroup and subscriptionId optional.

MSSedusch:
we found a major roadblock with the approach implemented in this fence agent. Hence closing the PR

oalbrigt:
New PR: https://github.com/ClusterLabs/fence-agents/pull/181

PR #192, opened 2018-04-04
use compat.py to provide compatibility for python2/3

liangxin1300:
hi:
In python3 environment, agent like fence_sbd which will call run_command function, will raise error about bytes and string convert

I think it's necessary to  use compatible code to handle this problem, 
we can append more functions to compat.py later

regards,
xin


oalbrigt:
Closing as #199 solved this issue.

PR #193, opened 2018-04-06
Low: fence_scsi: Remove period from cmd string

nrwahl2:
Changes `corosync-cmapctl nodelist.` to `corosync-cmapctl nodelist`.
Existing version works fine in my testing, so this is a simple typo
cleanup.


oalbrigt:
add to whitelist

oalbrigt:
Thanks.

PR #194, opened 2018-04-11
fence_gce: Clean up and log errors

krig:
Catch any exceptions and write them to the log.

Also generally make the code a bit more pythonic.

### DIFF
--- a/fence/agents/gce/fence_gce.py
+++ b/fence/agents/gce/fence_gce.py
@@ -3,11 +3,18 @@
 import atexit
 import sys
 sys.path.append("@FENCEAGENTSLIBDIR@")
-from googleapiclient import discovery
-from oauth2client.client import GoogleCredentials
 
-from fencing import *
-from fencing import fail, fail_usage, EC_TIMED_OUT, run_delay
+import googleapiclient
+import oauth2client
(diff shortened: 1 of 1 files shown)


oalbrigt:
I dont think we should rewrite tabs to spaces for indentation, as long as the indentation is consistent within the agent that should be good enough, and you might cause new issues by changing it in Python code.

krig:
alright, I’ll revert that.

oalbrigt:
LGTM.

fence/agents/gce/fence_gce.py
```diff
@@ -3,55 +3,57 @@
...
-				status = "on"
-			elif instance["status"] == "TERMINATED":
-				status = "off"
-			result[instance["id"]] = (instance["name"], status)
-	# TODO: check which Exceptions it can throw
-	except:
-		fail_usage("Failed: Unable to connect to GCE. Check your configuration.")
+			result[instance["id"]] = (instance["name"], translate_status(instance["status"]))
+	except Exception as err:
+		fail_usage("fence_gce.get_nodes_list: {}".format(str(err)))
```
oalbrigt: I think these fail_usage() calls should still say Failed: and fence_gce would be provided by the logging facility, so e.g. ""Failed: get_nodes_list: {}" in this case.

PR #197, opened 2018-04-20
fence_gce: google-auth instead of oauth2client

helen-fornazier:
oauth2client is deprecated, replacing it by google-auth

### DIFF
--- a/agents/gce/fence_gce.py
+++ b/agents/gce/fence_gce.py
@@ -5,7 +5,6 @@ import sys
 sys.path.append("@FENCEAGENTSLIBDIR@")
 
 import googleapiclient.discovery
-import oauth2client.client
 from fencing import fail_usage, run_delay, all_opt, atexit_handler, check_input, process_input, show_docs, fence_action
 
 def translate_status(instance_status):
@@ -99,7 +98,10 @@ def main():
 	run_delay(options)
 
 	try:
-		credentials = oauth2client.client.GoogleCredentials.get_application_default()
+		credentials = None
+		if tuple(googleapiclient.__version__) < tuple("1.6.0"):
+			import oauth2client.client
+			credentials = oauth2client.client.GoogleCredentials.get_application_default()
 		conn = googleapiclient.discovery.build('compute', 'v1', credentials=credentials)
 	except Exception as err:
 		fail_usage("Failed: Create GCE compute v1 connection: {}".format(str(err)))
--- a/agents/gce/fence_gce.py
+++ b/agents/gce/fence_gce.py
@@ -5,7 +5,7 @@ import sys
 sys.path.append("@FENCEAGENTSLIBDIR@")
 
 import googleapiclient.discovery
-import oauth2client
+import oauth2client.client
 from fencing import fail_usage, run_delay, all_opt, atexit_handler, check_input, process_input, show_docs, fence_action
 
 def translate_status(instance_status):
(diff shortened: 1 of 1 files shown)


fabbione:
@oalbrigt please let me know what packages are required in CI for this change to be tested properly.

fence/agents/gce/fence_gce.py
```diff
@@ -5,7 +5,8 @@
 sys.path.append("@FENCEAGENTSLIBDIR@")
 
 import googleapiclient
-import oauth2client
+import googleapiclient.discovery
```
oalbrigt: This is already imported in line 7.

fence/agents/gce/fence_gce.py
```diff
@@ -99,7 +100,7 @@ def main():
 	run_delay(options)
 
 	try:
-		credentials = oauth2client.GoogleCredentials.get_application_default()
+		credentials, _ = google.auth.default()
```
oalbrigt: This seems to not be available in a lot of distros yet, so I think we should move the import-line from line 9 to this try-statement, and have a fallback to use the old library (except ImportError:).

fence/agents/gce/fence_gce.py
helen-fornazier: importing the package doesn't mean it imported the module and I get an error that googleapiclient.discovery was not found. What I can do is to remove the line 7 and leave just the "import googleapiclient.discovery"

fence/agents/gce/fence_gce.py
helen-fornazier: Ack, I just noticed that googleapiclient already do that if I use the default credentials [1]

I am fixing this PR, sorry for the noise.

[1] https://github.com/google/google-api-python-client/blob/master/googleapiclient/_auth.py#L39

helen-fornazier:
@oalbrigt Hi, could you please take a look in this last version of the PR when you have some time please?

Thanks

helen-fornazier:
I removed the library requirements for google-auth. This requirement should be in googleapiclient, not in fences, so when googleapiclient updates, fences will use google-auth automatically.

googleapiclient is already migrating to google-auth, see:
https://github.com/google/google-api-python-client/pull/493
https://github.com/google/google-api-python-client/pull/499

oalbrigt:
Looks good to me. Does it also work fine with older versions without the credentials part?

oalbrigt:
add to whitelist

oalbrigt:
You'll have to run git rebase upstream/master and git push -f, as I've just merged a PR that restructures the source tree.

helen-fornazier:
> Looks good to me. Does it also work fine with older versions without the credentials part?
add to whitelist

It works from googleapiclient version 1.6.0, how can I add it to the whitelist? It doesn't seem that AC_PYTHON_MODULE accepts a version

helen-fornazier:
> I guess we could just use an except with whatever error it gives in versions < 1.6.0 where it uses the old method before the final exception, as we cant expect all distributions to use the latest versions.

I can't add a try/except there, because the old library only trows a login exception when it tries to use the conn object, not when building it.

I added a verification of the library version and I also added another commit fixing an import regression.
Please check if it is ok now, or if you want modifs, e.g. maybe I could move the check for the version and the import oauth2client.client to the top of the file.

helen-fornazier:
ops, sorry, I just resent fixing coding style, please check this last version of the PR.
Thanks

oalbrigt:
Great. That's much better. You can change the import googleapiclient back to what it was, as it's also required for the version-check now.

helen-fornazier:
The import works like this, I tested it, but if you want me to add the `import googleapiclient`, we will need to also keep the `import googleapiclient.discovery`, what do you prefer?

oalbrigt:
Oh. I didnt expect it to work when you imported sub parts of a library, so we're all good then.

krig:
Hey @helen-fornazier, apologies for breaking the imports!

helen-fornazier:
Hi @krig, no problem.

PR #202, opened 2018-05-06
Add agent for OVH Cloud (REST API)

cymen:
This adds an agent for the OVH Cloud servers via the REST API. The REST API uses Oauth for authentication. Because of the authentication method, the agent ends up expecting an application key, an application secret and a consumer key or token. Getting/creating these values is explained at [First Steps with the API](https://api.ovh.com/g934.first_step_with_api). When creating the consumer key or token, it's important to select the option that disables expiration of it. Noting this here as it was unclear where this kind of documentation should go.

I'm also a bit unclear on the metadata aspect however I've implemented both `on` and `off` (with polling to verify status before it exits with return value `0` or `1`).

### DIFF
--- a/agents/ovh_cloud_rest/fence_ovh_cloud_rest.py
+++ b/agents/ovh_cloud_rest/fence_ovh_cloud_rest.py
@@ -1,0 +1,124 @@
+#!@PYTHON@ -tt
+
+import sys
+import time
+import atexit
+sys.path.append("@FENCEAGENTSLIBDIR@")
+from fencing import *
+from fencing import fail_usage, run_command, run_delay
+import ovh
+
+POLL_INTERVAL_SECONDS = 1
+
+def define_new_opts():
+    all_opt["application_key"] = {
+            "getopt" : "k:",
+            "longopt" : "application-key",
+            "help" : "-k, --application-key=[application-key]               OVH Oauth application key",
+            "required" : "1",
+            "shortdesc" : "OVH Oauth application key",
+            "order" : 1}
+    all_opt["application_secret"] = {
+            "getopt" : "s:",
+            "longopt" : "application-secret",
+            "help" : "-s, --application-secret=[application-secret]         OVH Oauth application secret",
+            "required" : "1",
+            "shortdesc" : "OVH Oauth application secret",
+            "order" : 2}
+    all_opt["token"] = {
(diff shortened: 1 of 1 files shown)


cymen:
It's failing CI now on absence of `tests/data/metadata/fence_ovh_cloud_rest.xml`. Is that a hand created file or should the metadata part (that I mentioned being confused about above) be implemented and then this is generated?

agents/ovh_cloud_rest/fence_ovh_cloud_rest.py
```diff
@@ -0,0 +1,119 @@
...
+    docs["vendorurl"] = "https://api.ovh.com/"
+    show_docs(options, docs)
+
+    run_delay(options)
+
+    session = ovh_login(options);
+
+    rescue_url = "/cloud/project/%s/instance/%s/rescueMode" % (options['--service-name'], options['--instance-id'])
+
+    if options["--action"] == "off":
```
oalbrigt: You should put the on/off code into e.g. set_power_status() and make a get_power_status() where it returns on/off/unknown. This way it can e.g. send reboot as send power-off, wait for power off, send power-on, wait for power-on.

Similar agent with a good example on how to do this:
https://github.com/ClusterLabs/fence-agents/blob/master/agents/vmware_rest/fence_vmware_rest.py#L178

oalbrigt:
You can run "make xml-upload" to generate/update this.

agents/ovh_cloud_rest/fence_ovh_cloud_rest.py
cymen: Ah, good idea. I misread part of the `FenceAgentAPI` doc and thought it meant reboot was a bad idea in general (but I see it suggests that only that is a bad idea for fabric fencing).

I'll make this change.

cymen: @oalbrigt One question -- does it matter if my `get_power_status` fails? As in, will the code in `fencing.py` retry until it gets a valid status? Or does my code need to sleep and poll until it gets a valid power status (as I've done in the code here)? It's not 100% obvious to me as I don't know if the vmware example one sometimes returns states between off and on as the OVH one does (ie if going to RESCUE, it returns RESCUING until it is ready and then RESCUE).

Issue #206, opened 2018-05-22
fence_scsi: not Python3 safe


bootc:
Using `fence_scsi` with python3 as the interpreter leads to errors like this:

```2018-05-22 14:44:19,110 INFO: Executing: /usr/sbin/corosync-cmapctl totem.cluster_name
2018-05-22 14:44:19,118 DEBUG: 0 totem.cluster_name (str) = zfs
Traceback (most recent call last):
  File "/usr/sbin/fence_scsi", line 512, in <module>
    main()
  File "/usr/sbin/fence_scsi", line 489, in main
    options["--key"] = generate_key(options)
  File "/usr/sbin/fence_scsi", line 199, in generate_key
    return "%.4s%.4d" % (get_cluster_id(options), int(get_node_id(options)))
  File "/usr/sbin/fence_scsi", line 184, in get_cluster_id
    return hashlib.md5(match.group(1)).hexdigest()
TypeError: Unicode-objects must be encoded before hashing
```

oalbrigt:
What version of fence-agents are you running?

bootc:
This was using Debian's 4.2.0-1. I've also reported the bug there: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=899310

oalbrigt:
Can you update "match.group(1)" to "match.group(1).encode()" and see if that solves the issue for you?

bootc:
Yes the patch on the Debian bug does that and appears to fix it:
https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=899310;filename=fence_scsi.patch;msg=15

oalbrigt:
Thanks. I've applied that patch in this PR: https://github.com/ClusterLabs/fence-agents/pull/208

oalbrigt:
Adding your patch for reference, as I closed mine: https://github.com/ClusterLabs/fence-agents/pull/207

PR #207, opened 2018-05-23
fence_scsi: fix python3 encoding error #206

vvidic:
File "/usr/sbin/fence_scsi", line 184, in get_cluster_id
  return hashlib.md5(match.group(1)).hexdigest()
TypeError: Unicode-objects must be encoded before hashing


oalbrigt:
add to whitelist

oalbrigt:
Thanks.

PR #211, opened 2018-06-26
fence_gce: logging and non required parameters

helen-fornazier:
The following changes were made provide some of the features supported by the following script:  https://storage.googleapis.com/sapdeploy/pacemaker-gcp/gcpstonith

fence_gce: add support for stackdriver logging
Add --logging option to enable sending logs to google stackdriver

fence_gce: set project and zone as not required

Try to retrieve the GCE project if the script is being executed inside a
GCE machine if --project is not provided.
Try to retrieve the zone automatically from GCE if --zone is not
provided.

### DIFF
--- a/agents/gce/fence_gce.py
+++ b/agents/gce/fence_gce.py
@@ -83,6 +83,21 @@ def set_power_status(conn, options):
 		fail_usage("Failed: set_power_status: {}".format(str(err)))
 
 
+def power_cycle(conn, options):
+	try:
+		LOGGER.info('Issuing reset of %s in zone %s' % (options["--plug"], options["--zone"]))
+		operation = conn.instances().reset(
+				project=options["--project"],
+				zone=options["--zone"],
+				instance=options["--plug"]).execute()
+		wait_for_operation(conn, options["--project"], options["--zone"], operation)
+		LOGGER.info('Reset of %s in zone %s complete' % (options["--plug"], options["--zone"]))
+		return True
+	except Exception as err:
+		LOGGER.error("Failed: power_cycle: {}".format(str(err)))
+		return False
+
+
 def get_instance(conn, project, zone, instance):
 	request = conn.instances().get(
 			project=project, zone=zone, instance=instance)
@@ -161,13 +176,15 @@ def main():
 
 	hostname = platform.node()
 
-	device_opt = ["port", "no_password", "zone", "project", "logging"]
+	device_opt = ["port", "no_password", "zone", "project", "logging", "method"]
 
 	atexit.register(atexit_handler)
 
 	define_new_opts()
 
 	all_opt["power_timeout"]["default"] = "60"
+	all_opt["method"]["default"] = "cycle"
+	all_opt["method"]["help"] = "-m, --method=[method]          Method to fence (onoff|cycle) (Default: cycle)"
 
 	options = check_input(device_opt, process_input(device_opt))
 
@@ -225,7 +242,7 @@ def main():
 			fail_usage("Failed retrieving GCE zone. Please provide --zone option: {}".format(str(err)))
 
 	# Operate the fencing device
-	result = fence_action(conn, options, set_power_status, get_power_status, get_nodes_list)
+	result = fence_action(conn, options, set_power_status, get_power_status, get_nodes_list, power_cycle)
 	sys.exit(result)
 
 if __name__ == "__main__":
--- a/agents/gce/fence_gce.py
+++ b/agents/gce/fence_gce.py
@@ -2,9 +2,18 @@
 
 import atexit
 import logging
+import os
 import platform
 import sys
 import time
+if sys.version_info >= (3, 0):
+  # Python 3 imports.
+  import urllib.parse as urlparse
+  import urllib.request as urlrequest
+else:
+  # Python 2 imports.
+  import urllib as urlparse
+  import urllib2 as urlrequest
 sys.path.append("@FENCEAGENTSLIBDIR@")
 
 import googleapiclient.discovery
--- a/agents/gce/fence_gce.py
+++ b/agents/gce/fence_gce.py
@@ -20,7 +20,6 @@ import googleapiclient.discovery
 from fencing import fail_usage, run_delay, all_opt, atexit_handler, check_input, process_input, show_docs, fence_action
 
 
-LOGGER = logging
 METADATA_SERVER = 'http://metadata.google.internal/computeMetadata/v1/'
 METADATA_HEADERS = {'Metadata-Flavor': 'Google'}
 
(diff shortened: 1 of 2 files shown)


oalbrigt:
You should use the built-in logging mechanisms in the fencing library.

helen-fornazier:
right, I'll do a wrapper then, because I need to send the log to the GCE stackdriver

oalbrigt:
Yeah. It supports -v to show logs on console, and -D to log to file. Btw. You'll have to run make xml-upload" and add the updated metadata to the PR as well to avoid CI failing.

helen-fornazier:
Ah yes, sorry about that, I'll update the PR

oalbrigt:
No worries.

helen-fornazier:
@oalbrigt I updated the PR, could you take a look please?

oalbrigt:
Could you add it to the default logger, so we can do all the logging via logging.*()?
https://github.com/ClusterLabs/fence-agents/blob/master/lib/fencing.py.py#L693

helen-fornazier:
I am not sure I understand what you mean, I am using the default logger https://github.com/ClusterLabs/fence-agents/pull/211/commits/6b7fef09f556613707edad20295ce59fc607afd0#diff-4a0f3ce2f5196c812280cc52f7aa42b6R25

Do you mean to add a handler instead? To wrap the default behavior?

oalbrigt:
Yeah.

helen-fornazier:
@oalbrigt I updated the PR, could you please check now?
I also added another commit implementing power cycle, let me know if you prefer a different PR for that.

Thanks

helen-fornazier:
@oalbrigt fyi, I created a child Logger to avoid every message to be sent to the new handler, but as it is a child logger, it will call the default parent logging.* as well.

oalbrigt:
This is done by removing the : in the getopt line: "getopt" : "",

helen-fornazier:
> This is done by removing the : in the getopt line: "getopt" : "",

This works great, thanks

> I think you should add it to the main logger, as it makes sense to log these to where users look for their logs in general too (without the -v parameter only WARNING or higher are logged, so it wont "spam" the logs).

The problem is that google python lib prints a warning for every execution about a feature that won't be used (and is not necessary to function), which floods the stackdriver logs.
`WARNING: file_cache is unavailable when using oauth2client >= 4.0.0`

I was also reading this blog post: https://www.electricmonk.nl/log/2017/08/06/understanding-pythons-logging-module/
Which says it is better to use a child logger then the root logger to have more control about the logs.

oalbrigt:
Thanks. I'll see if I can find a workaround for that.

oalbrigt:
You can suppress those Warnings with the following line:
logging.getLogger("googleapiclient.discovery_cache").setLevel(logging.ERROR)

helen-fornazier:
> You can suppress those Warnings with the following line:
> logging.getLogger("googleapiclient.discovery_cache").setLevel(logging.ERROR)

I don't like this, because it means we are suppressing those messages in the root logger too, so it won't show up when executing with -v or -D

helen-fornazier:
> > You can suppress those Warnings with the following line:
> > logging.getLogger("googleapiclient.discovery_cache").setLevel(logging.ERROR)
> 
> I don't like this, because it means we are suppressing those messages in the root logger too, so it won't show up when executing with -v

Never mind, I am find with this, please check my last commit

agents/gce/fence_gce.py
```diff
@@ -38,50 +56,141 @@ def get_power_status(conn, options):
...
+		"required" : "0",
+		"order" : 4
+	}
+
 
 def main():
 	conn = None
 
-	device_opt = ["port", "no_password", "zone", "project"]
+	hostname = platform.node()
```
oalbrigt: You should use options["--plug"] instead of this, to avoid incosistency (and users might use different instance name and hostname).

oalbrigt:
Looks good to me.

By the way. Is the reset-function guaranteed to be hard-reset? Some systems use soft-reset (pretty much like running "reboot" on the node), which isnt good for fencing, so we should only use cycle as default if it's always hard-reset.

helen-fornazier:
It is, from https://cloud.google.com/compute/docs/instances/restarting-an-instance

    Performing a reset on your instance is similar to pressing the reset button on your computer

oalbrigt:
Great.

agents/gce/fence_gce.py
helen-fornazier: hostname is just for logging to stackdriver

oalbrigt: Yeah, but I dont think it's necessary to import platform and define a new variable here when you can just use options["--plug"] on line 214 instead.

agents/gce/fence_gce.py
```diff
@@ -97,6 +206,25 @@ def main():
...
+		try:
+			import google.cloud.logging.handlers
+			client = google.cloud.logging.Client()
+			handler = google.cloud.logging.handlers.CloudLoggingHandler(client, name=hostname)
+			handler.setLevel(logging.INFO)
+			formatter = logging.Formatter('gcp:stonish "%(message)s"')
+			handler.setFormatter(formatter)
+			root_logger = logging.getLogger()
+			if options.get('--verbose') is None:
+				root_logger.setLevel(logging.INFO)
```
oalbrigt: This doesnt make any sense. The logic for the root_logger is taken care of in fencing.py.

agents/gce/fence_gce.py
```diff
@@ -97,6 +206,25 @@ def main():
...
+				logging.getLogger("googleapiclient").setLevel(logging.ERROR)
```
oalbrigt: This should be put outside stackdriver-logging, as we dont want to see these warnings for every run.

oalbrigt:
@helen-fornazier Pacemaker already does that for us, so that's not necessary (telling which fence device is used in the log before the messages).

By the way it's STONITH, not STONISH.

agents/gce/fence_gce.py
helen-fornazier: hostname is only used by stackdriver logs, if I understood correctly, the action `-o list` does't require `--plug`, so if I remove the hostname, the stackdriver won't be initialized correctly if no `--plug` is provided.

agents/gce/fence_gce.py
helen-fornazier: yes, but if we use `--stackdriver-logging`, the logs are not being sent to stackdriver because the logging level defaults to WARN and not INFO.

agents/gce/fence_gce.py
helen-fornazier: ok, I'll update the PR.

agents/gce/fence_gce.py
oalbrigt: Cant you just add the handler above (which has the correct level for stackdriver logging) like we do in the fencing library?

Example:
https://github.com/ClusterLabs/fence-agents/blob/master/lib/fencing.py.py#L723

agents/gce/fence_gce.py
oalbrigt: I guess you could skip stackdriver logging for those calls (as local logging should be sufficient when not fencing or checking the status of an instance).

agents/gce/fence_gce.py
helen-fornazier: I am not sure I understand what you mean, I added the handler here: https://github.com/ClusterLabs/fence-agents/pull/211/files/9ae0a072424fa982e1d18a2cb661628c38601c3a#diff-4a0f3ce2f5196c812280cc52f7aa42b6R222

agents/gce/fence_gce.py
helen-fornazier: agreed, I'll update the PR

agents/gce/fence_gce.py
oalbrigt: Oh. I didnt find that. But doesnt that log correctly based on the handler.setLevel() above?

helen-fornazier: no, this is just a filter to the handler. The root logger get the logs based in the root logger level, then it sends the logs to the handlers which filter the messages again with its own level, but if the logs does't get thought the root logger in the first place, then nothing is sent to the handlers.

oalbrigt: Ah. I guess we'll just have to live with that, and I might improve that later by lowering the other sub-loggers in this case.

helen-fornazier: This is a nice chart flow explaining that: https://docs.python.org/3/howto/logging.html#logging-flow

oalbrigt: Thanks.

helen-fornazier:
@oalbrigt I the changes we agreed on in this last commit, could you please check now? I just wans't sure what to do about setting the level of the logging, so I left it as it was, please check.

Thanks for reviewing this.

oalbrigt:
Looks good to me now. And I think the logging is good too.

oalbrigt:
Thanks.

PR #214, opened 2018-07-01
[build] be more aggressive in cleaning the source tree

fabbione:
Signed-off-by: Fabio M. Di Nitto <fdinitto@redhat.com>


fabbione:
Spoke to @oalbrigt on google hangout and he agreed this should be merged.

PR #218, opened 2018-07-19
fence_gce: filter call to aggregatedList

helen-fornazier:
Don't list all the instances in the project, filter only the one we are
interested in.


helen-fornazier:
@oalbrigt fyi, this is basically the sabe PR as this one: https://github.com/ClusterLabs/resource-agents/pull/1176
Please, let me know if this is ok

Issue #224, opened 2018-08-08
incorrect help for fence_rhevm -n (--plug) option


emesika:
fence_rhevm --help  :

-n, --plug=[id]                Physical plug number on device, UUID or
                                        identification of machine

This is not correct, the parameter must map to the VM name as known to RHV

Should be

 -n, --plug=[id]                The VM name in RHV

PR #229, opened 2018-08-14
The help for the fence_rhevm -n (--plug) parameter is incorrect.

emesika:
The help for the fence_rhevm -n (--plug) parameter is incorrect.

The current help is :

 -n, --plug=[id]                Physical plug number on device, UUID or
                                        identification of machine
should be

-n, --plug=[name]		The VM name in RHV

The fence_rhevm using the VM name passed in order to get the UUID of
that VM such that the VM can be stopped/started user REST API

<url>/vms/<vm uuid>/stop or <url>/vms/<vm uuid>/start

Issue : https://github.com/ClusterLabs/fence-agents/issues/224
Signed-off-by: emesika <emesika@redhat.com>

### DIFF
--- a/agents/rhevm/fence_rhevm.py
+++ b/agents/rhevm/fence_rhevm.py
@@ -136,6 +136,14 @@ def send_command(opt, command, method="GET"):
 	return result
 
 def define_new_opts():
+
+	all_opt["port"] = {
+		"getopt" : "n:",
+		"longopt" : "plug",
+		"help" : "-n, --plug=[name]              "
+		"The VM name in RHV",
+		"required" : "1",
+		"order" : 1}
 	all_opt["use_cookies"] = {
 		"getopt" : "",
 		"longopt" : "use-cookies",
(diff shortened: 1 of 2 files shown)


oalbrigt:
This is a generic description, and the last part of it seems to match your new description.

emesika:
regarding https://github.com/ClusterLabs/fence-agents/pull/229#issuecomment-412855629
>> This is a generic description, and the last part of it seems to match your new description.

That's not true :+1: 
n, --plug=[id] Physical plug number on device, UUID or
identification of machine

The only acceptable description for the plug parameter in the case of fence_rhevm is that it shoM name , UUID of the VM will not work

oalbrigt:
"identification of machine" is the VM name in this case, so it's already there.

mwperina:
> "identification of machine" is the VM name in this case, so it's already there.

VM can be uniquely identified in oVirt/RHV by UUID, VM name or FQDN, but only VM name works in the case of fence_rhevm. So, Oyvind, could you please merge it to make this help message really clear for users? We have quite a few questions about it from various users what they need to supply as a parameter here ...

@oalbrigt

Issue #225, opened 2018-08-08
fence_rhevm does not return the correct error in case of failure 


emesika:
fence_rhevm uses REST API  to call RHV 
In case that the call fails an unrelated error is displayed 

"Failed: Unable to obtain correct plug status or plug is not available"

Code should be changed to output the error returned from the REST API call

PR #235, opened 2018-08-27
fence-rhevm: improve error reporting.

emesika:
fence_rhevm uses REST API to call RHV

In fence_rhevm , the plug parameter is actually the VN name and is used
in order to construct a REST API search request to get the VM UUID from
RHV by the VM name.
In case that the call fails an unrelated error is displayed

"Failed: Unable to obtain correct plug status or plug is not available"

Code should be changed to output a clear error in case that the REST API fail

Issue : #225
Signed-off-by: emesika emesika@redhat.com


emesika:
@oalbrigt can you please review (https://github.com/oalbrigt)

oalbrigt:
Thanks

PR #244, opened 2018-10-12
lib: Add support for action validate-all-xml

marxsk:
Output of this action is machine-readable format in the XML format that
can be used for processing in pacemaker/pcs/UI

### DIFF
--- a/lib/fencing.py.py
+++ b/lib/fencing.py.py
@@ -663,7 +663,7 @@ def process_input(avail_opt):
 ## in each of the fencing agents. It looks for possible errors and run
 ## password script to set a correct password
 ######
-def check_input(device_opt, opt, other_conditions = False):
+def check_input(device_opt, opt, local_validation = None):
 	device_opt.extend(_add_dependency_options(device_opt))
 
 	options = dict(opt)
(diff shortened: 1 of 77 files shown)


marxsk:
@oalbrigt 
It will be required to update XML metadata as well but I did not want to enlarge PR yet. 

Please, take a look at code and possible extensions also from the point of resource-agents

lib/fencing.py.py
```diff
@@ -711,10 +711,16 @@ def check_input(device_opt, opt, other_conditions = False):
 		options["--action"] = "off"
 
 
-	if options["--action"] == "validate-all" and not other_conditions:
-		if not _validate_input(options, False):
```
oalbrigt: I think it'd be better to make a variable called e.g. validate_format to use here, so we can skip the elif part below.

lib/fencing.py.py
```diff
@@ -1295,64 +1301,130 @@ def _set_default_values(options):
...
+		})
 
 	if device_opt.count("ipaddr") and "--ip" not in options and "--managed" not in options and "--target" not in options:
-		valid_input = False
-		fail_usage("Failed: You have to enter fence address", stop)
+		errors.append({
+			'fields': ['ipaddr', 'managed', 'target'],
+			'text': 'You have to enter fence address',
+			'text_cli': 'You have to enter fence address (-a)',
+			'error_type': 'REQUIRE',
```
oalbrigt: Is this one-or-more, vs one-of being one or the other?

lib/fencing.py.py
```diff
@@ -711,10 +711,16 @@ def check_input(device_opt, opt, other_conditions = False):
...
 
 
-	if options["--action"] == "validate-all" and not other_conditions:
-		if not _validate_input(options, False):
-			fail_usage("validate-all failed")
-		sys.exit(EC_OK)
+	if options["--action"] in ["validate-all", "validate-all-xml"] and not other_conditions:
+		if options["--action"] == "validate-all":
+			if not _validate_input(options, False, 'text'):
+				fail_usage("validate-all failed")
```
marxsk: add sys.exit(EC_GENERIC_ERROR)

lib/fencing.py.py
marxsk: I like the idea of having validate_actions. But it won't help us with if/elif part, the code still have to be branched.

lib/fencing.py.py
marxsk: one-or-more; which name do you suggest? ONE-OR-MORE?

oalbrigt:
add to whitelist

oalbrigt:
add to whitelist

oalbrigt:
retest this please

fabbione:
@marxsk  apparently github has closed this issue when renaming master branch to main. I am unable to reopen the PR for you unfortunately, as you need to change the target branch from master to main.

PR #247, opened 2018-10-25
fence_openstack: Migrate from keystoneclient to keystoneauth1

racciari:
keystoneclient is deprecated in favor of keystoneauth1.

### DIFF
--- a/agents/openstack/fence_openstack.py
+++ b/agents/openstack/fence_openstack.py
@@ -11,12 +11,15 @@ from fencing import *
 from fencing import fail_usage, is_executable, run_command, run_delay
 
 try:
-        from keystoneclient.v3 import client as ksclient
+        from novaclient import client as novaclient
+        from keystoneauth1 import session as ksc_session
+        from keystoneauth1 import loading
+        legacy_import = False
+except ImportError:
         from novaclient import client as novaclient
         from keystoneclient import session as ksc_session
         from keystoneclient.auth.identity import v3
-except ImportError:
-        pass
+        legacy_import = True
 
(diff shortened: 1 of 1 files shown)


oalbrigt:
Can you add fallback to keystoneclient, so it still works for systems with keystoneclient?

racciari:
Here you go, but I've not tested the legacy import with keystoneclient.

agents/openstack/fence_openstack.py
```diff
@@ -11,12 +11,15 @@
...
 try:
-        from keystoneclient.v3 import client as ksclient
+        from novaclient import client as novaclient
+        from keystoneauth1 import session as ksc_session
+        from keystoneauth1 import loading
+        legacy_import = False
+except ImportError:
         from novaclient import client as novaclient
         from keystoneclient import session as ksc_session
         from keystoneclient.auth.identity import v3
```
oalbrigt: I think you meant to put `legacy_import=True` here.

racciari: The `legacy_import=True` is in the `ImportError` section below

oalbrigt:
ok to test

agents/openstack/fence_openstack.py
oalbrigt: It needs some extra nesting, so I'll take care of that after merging.

oalbrigt:
Created https://github.com/ClusterLabs/fence-agents/pull/248 to add the except: pass back for metadata generation, and also added keystonauth1 to our library detection.

PR #250, opened 2018-11-16
Fix 'log_expect' in fence_hpblade.py

mmartinv:
Update the 'log_expect' call to the new method definition.


oalbrigt:
LGTM. Thanks.

PR #253, opened 2018-12-03
Add new fence agent for Redfish

jose-delarosa:
- Agent works on all fence devices that implement the Redfish API specification
- Agent programatically finds the Systems Resouce URI if it's not provided

### DIFF
--- a/agents/redfish/fence_redfish.py
+++ b/agents/redfish/fence_redfish.py
@@ -117,7 +117,7 @@ def main():
 
     opt = process_input(device_opt)
 
-    all_opt["ipport"]["default"] = "443"
+    all_opt["ssl"]["default"] = "1"
     options = check_input(device_opt, opt)
 
     docs = {}
--- a/agents/redfish/fence_redfish.py
+++ b/agents/redfish/fence_redfish.py
@@ -13,7 +13,7 @@ sys.path.append("@FENCEAGENTSLIBDIR@")
 
 from requests.packages.urllib3.exceptions import InsecureRequestWarning
 from fencing import *
-from fencing import fail_usage
+from fencing import fail_usage, run_delay
 
 def get_power_status(conn, options):
     uri = options["--systems-uri"]
@@ -127,6 +127,7 @@ Out-of-Band controllers that support Redfish APIs. These controllers provide rem
 access to control power on a server."
     docs["vendorurl"] = "http://www.dmtf.org"
     show_docs(options, docs)
+    run_delay(options)
 
     ##
     ## Operate the fencing device
--- a/agents/redfish/fence_redfish.py
+++ b/agents/redfish/fence_redfish.py
@@ -1,0 +1,151 @@
+#!@PYTHON@ -tt
+
+# Copyright (c) 2018 Dell Inc. or its subsidiaries. All Rights Reserved.
+
+# Fence agent for devices that support the Redfish API Specification.
+
+import sys
+import re
+import json
+import requests
+import atexit
+sys.path.append("@FENCEAGENTSLIBDIR@")
+
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
+from fencing import *
+from fencing import fail_usage
+
+def get_power_status(conn, options):
+    uri = options["--systems-uri"]
+    response = send_get_request(options, uri)
+    if response['ret'] is False:
+        fail_usage("Couldn't get power information")
+    data = response['data']
+    if data[u'PowerState'].strip() == "On":
+        return "on"
+    else:
+        return "off"
+
+def set_power_status(conn, options):
+    action = {
+        'on' : "On",
+        'off': "ForceOff",
+        'reboot': "GracefulRestart"
+    }[options["--action"]]
+
+    payload = {'ResetType': action}
+    headers = {'content-type': 'application/json'}
+
+    # Search for 'Actions' key and extract URI from it
+    uri = options["--systems-uri"]
+    response = send_get_request(options, uri)
+    if response['ret'] is False:
+        return {'ret': False}
+    data = response['data']
+    uri = data["Actions"]["#ComputerSystem.Reset"]["target"]
+
+    response = send_post_request(options, uri, payload, headers)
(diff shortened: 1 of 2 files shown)


oalbrigt:
Can you run `make xml-upload` and add the updated metadata to this PR?

jose-delarosa:
Done, although I still see an error with `make -j4 check`.

oalbrigt:
Ah. Now it fails due to the delay-check somehow taking longer or shorter than expected.

jose-delarosa:
Please let me know if you have any questions, thanks.

agents/redfish/fence_redfish.py
```diff
@@ -0,0 +1,152 @@
...
+def get_power_status(conn, options):
+    uri = options["--systems-uri"]
+    response = send_get_request(options, uri)
+    if response['ret'] is False:
+        fail_usage("Couldn't get power information")
+    data = response['data']
+    if data[u'PowerState'].strip() == "On":
+        return "on"
+    else:
+        return "off"
```
oalbrigt: Are there any other states? We only want to return off when it's completely off, and you can return "unknown" for the other states so it retries until it's actually off.

oalbrigt:
ok to test

agents/redfish/fence_redfish.py
jose-delarosa: Good point. There are 2 other states that are very brief (but still possible): PoweringOn and PoweringOff, and of course Off. I will update this code to return "off" only when the state is "Off".

jose-delarosa: Done

agents/redfish/fence_redfish.py
```diff
@@ -0,0 +1,152 @@
...
+    if data[u'PowerState'].strip() == "Off":
+        return "off"
+    else:
+        return "on"
+
+def set_power_status(conn, options):
+    action = {
+        'on' : "On",
+        'off': "ForceOff",
+        'reboot': "GracefulRestart"
```
oalbrigt: Is there a hard-reset version we can use instead of GracefulRestart?

GracefulRestart might get stuck on e.g. I/O or systemd/init-scripts stopping processes, which will make the boot process take a long time or "forever" in worst-case scenario.

agents/redfish/fence_redfish.py
```diff
@@ -0,0 +1,152 @@
...
+import atexit
+sys.path.append("@FENCEAGENTSLIBDIR@")
+
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
+from fencing import *
+from fencing import fail_usage, run_delay
+
+def get_power_status(conn, options):
+    uri = options["--systems-uri"]
+    response = send_get_request(options, uri)
```
marxsk: uri variable is use just once and it will be readable with options["..."] as well. This is there multiple times but it has no real impact, so don't fix if you like it that way more.

agents/redfish/fence_redfish.py
```diff
@@ -0,0 +1,152 @@
...
+    payload = {'ResetType': action}
+    headers = {'content-type': 'application/json'}
+
+    # Search for 'Actions' key and extract URI from it
+    uri = options["--systems-uri"]
+    response = send_get_request(options, uri)
+    if response['ret'] is False:
+        return {'ret': False}
+    data = response['data']
+    uri = data["Actions"]["#ComputerSystem.Reset"]["target"]
```
marxsk: I would prefer to not reuse variables

agents/redfish/fence_redfish.py
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+def main():
+    atexit.register(atexit_handler)
+    device_opt = ["ipaddr", "login", "passwd", "redfish-uri", "systems-uri", "ssl"]
+    define_new_opts()
+
+    opt = process_input(device_opt)
+
+    all_opt["ipport"]["default"] = "443"
```
marxsk: If you have to use SSL - it is better to set it directly. When ssl/ssl-secure/ssl-insecure is used than ipport's default is set automatically at run-time.

agents/redfish/fence_redfish.py
jose-delarosa: Yes there is a ForceRestart option. I will add it.

agents/redfish/fence_redfish.py
jose-delarosa: Agree, will change.

agents/redfish/fence_redfish.py
jose-delarosa: I will take a look. I prefer simpler code myself :)

agents/redfish/fence_redfish.py
marxsk: all_opt["ssl"]["default"] = "1" should be enough - you can easily check if it works in -o metadata

agents/redfish/fence_redfish.py
jose-delarosa: Only way I could simplify this is with:

`response = send_get_request(options, options["--systems-uri"])`

There is one instance when I call `send_get_request` where uri is not a part of options, so I need to pass a second parameter to specify the URI to use.

jose-delarosa: If it's OK with you, left this as-is.

oalbrigt: My only suggestion here is that you could just use `options["--systems-uri"]` in place of uri, as uri is only used once the places I've seen it defined in the code.

jose-delarosa: Done. Thank you @oalbrigt and @marxsk for your review. Let me know if you have other comments.

oalbrigt:
retest this please

oalbrigt:
Looks good to me now. My only concern is if we should should return "unknown" between the on and off states (we usually do).

What do you think @marxsk ?

marxsk:
Unknown is more readable but our main idea that we have to be sure with 'off' and rest does not matter that much is used. Fine for me

oalbrigt:
Thank you for the agent, and improvements.

PR #255, opened 2019-01-29
fence_rhevm: Changed Encoding to UTF-8

Numblesix:
Starting from RHV/Ovirt 4.2 we saw issues with the agent(unable to fence) after switching to UTF-8 all worked again.

### DIFF
--- a/agents/rhevm/fence_rhevm.py
+++ b/agents/rhevm/fence_rhevm.py
@@ -88,7 +88,7 @@ def send_command(opt, command, method="GET"):
 	## send command through pycurl
 	conn = pycurl.Curl()
 	web_buffer = io.BytesIO()
-	conn.setopt(pycurl.URL, url.encode("ascii"))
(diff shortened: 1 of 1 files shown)


oalbrigt:
Are you using another on your system? I just want to know what causes the issue in case we might need to do something more advanced to avoid it for all in the future.

Numblesix:
I tested it on 4 fresh RHEL7.6 installation all on de_DE.UTF-8, maybe as a side note we only saw this when the pcs cluster tried to fence(stonith admin) the other node in the cluster (2 node Cluster)

oalbrigt:
Thanks. That confirms my worries. I think we'll have to default to UTF-8, and have an option to change it, that I'll add.

Numblesix:
Why would somebody need to change it from UTF-8 ? Just to understand it :)

oalbrigt:
Maybe if Japanese or Chinese use other charsets.

PR #261, opened 2019-02-27
fence_pve: add support for sending reset command to qemu machines

rhl-jfm:
Restarting VMs by sending poweroff and poweron commands is slow and unreliable; machines may hang or time out during shutdown.

A better option, at least for qemu VMs, is to send a "reset" command. This immediately reboots the VM, just like pressing the reset button on a physical server.

This patch implements the required changes. The new reboot code is activated by providing the option method=cycle.

### DIFF
--- a/agents/pve/fence_pve.py
+++ b/agents/pve/fence_pve.py
@@ -56,7 +56,7 @@ def set_power_status(conn, options):
 
 def reboot_cycle(conn, options):
 	del conn
-	cmd = "nodes/" + options["--nodename"] + "/qemu/" + options["--plug"] + "/status/reset"
+	cmd = "nodes/" + options["--nodename"] + "/" + options["--vmtype"] + "/" + options["--plug"] + "/status/reset"
 	result = send_cmd(options, cmd, post={"skiplock":1})
 	return type(result) is dict and "data" in result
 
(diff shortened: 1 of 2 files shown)


agents/pve/fence_pve.py
```diff
@@ -54,6 +54,13 @@ def set_power_status(conn, options):
 	send_cmd(options, cmd, post={"skiplock":1})
 
 
+def reboot_cycle(conn, options):
+	del conn
+	cmd = "nodes/" + options["--nodename"] + "/qemu/" + options["--plug"] + "/status/reset"
```
oalbrigt: You should use `options["--vmtype"]` like in `set_power_status()` instead of `/qemu/`, so it can be used for all supported vmtypes.

oalbrigt:
ok to test

agents/pve/fence_pve.py
rhl-jfm: It's a fair point, and I can certainly make the change. It won't make much difference, though, since the reset action is only available for the qemu machine type anyway.

Ideally, fence_pve should notice the mismatch and gracefully fall back to the onoff method for lxc vmtypes.

oalbrigt: Oh. I can add some code to do that.

oalbrigt:
retest this please

PR #265, opened 2019-04-15
fence_redfish: add headers to HTTP methods

jose-delarosa:
* Needed for full compliance with Redfish spec.
* May cause errors in some devices if not sent.

### DIFF
--- a/agents/redfish/fence_redfish.py
+++ b/agents/redfish/fence_redfish.py
@@ -16,6 +16,11 @@ from requests.packages.urllib3.exceptions import InsecureRequestWarning
 from fencing import *
 from fencing import fail_usage, run_delay
 
+GET_HEADERS = {'accept': 'application/json', 'OData-Version': '4.0'}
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

oalbrigt:
Have you seen any issues with fencing not working due to this, or do you just get errors in the device log?

jose-delarosa:
@oalbrigt I have seen errors in some devices where the fencing operation fails (HTTP Error 406). Not every device enforces the presence of these headers, but some do. Adding these headers makes it compliant with the Redfish spec (see [here](https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.6.1.pdf), page 26).

oalbrigt:
Thanks for the information. That helps prioritize adding it in distros higher than if it was just causing error messages.

PR #266, opened 2019-04-16
fence_ipmilan: Add ipmitool timeout option

krig:
The default ipmitool timeout in some versions is
extremely short (on the order of 1s), and in
newer versions it is 2s by default.

This commit adds an option to configure the
timeout, and sets the default timeout to 2s.

Signed-off-by: Kristoffer Gronlund <kgronlund@suse.com>


oalbrigt:
LGTM. Thanks.

PR #279, opened 2019-07-11
Update fence_aliyun.py

dongchen126:
1.Import region provider to enable Alibaba cloud private zone functionality which support  API endpoint access from intranet
2.Add ram role functionality to improve security which disable AccessKey ID and Access Key Secret

### DIFF
--- a/agents/aliyun/fence_aliyun.py
+++ b/agents/aliyun/fence_aliyun.py
@@ -123,9 +123,9 @@ def define_new_opts():
 		"order" : 4
 	}
 	all_opt["ram_role"] = {
-		"getopt": "m:",
+		"getopt": ":",
 		"longopt": "ram-role",
-		"help": "-m, --ram-role=[name]        Ram Role",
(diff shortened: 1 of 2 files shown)


dongchen126:
1.Import region provider to enable Alibaba cloud private zone functionality which support API endpoint access from intranet
2.Add ram role functionality to improve security which disable AccessKey ID and Access Key Secret

agents/aliyun/fence_aliyun.py
```diff
@@ -121,12 +122,20 @@ def define_new_opts():
 		"required" : "0",
 		"order" : 4
 	}
+	all_opt["ram_role"] = {
+		"getopt": "m:",
```
oalbrigt: Can you update this to `":"`, and remove `-m` from the help text? I dont think it makes sense having `-m` for the `ram-role`.

oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #281, opened 2019-07-25
replace hard-to-read formatting marks with quotes where appropriate

OndrejHome:
This is done to improve readability of text when reading it from output of '-o metadata' like 'pcs' does for only 2 agents where these formatting string were found.

~~~
$ sed -i "s:\\\\\\\\f[IPRB]:':g" ./agents/kdump/fence_kdump.c
$ sed -i "s:\\\\\\\\f[IPRB]:':g" ./agents/vmware/fence_vmware.py
~~~

Text that was defining command line arguments has no quotes, just removed formatting strings.

If anything else needs changing in here feel free to let me know. This should not affect functionality. This targets only documentation of the `fence_kdump` and `fence_vmware` fence agents.

### DIFF
--- a/agents/kdump/fence_kdump.c
+++ b/agents/kdump/fence_kdump.c
@@ -206,12 +206,12 @@ do_action_metadata (const char *self)
     fprintf (stdout, "<resource-agent name=\"%s\"", basename (self));
     fprintf (stdout, " shortdesc=\"fencing agent for use with kdump crash recovery service\">\n");
     fprintf (stdout, "<longdesc>");
-    fprintf (stdout, "\\fIfence_kdump\\fP is an I/O fencing agent to be used with the kdump\n"
-                     "crash recovery service. When the \\fIfence_kdump\\fP agent is invoked,\n"
+    fprintf (stdout, "fence_kdump is an I/O fencing agent to be used with the kdump\n"
+                     "crash recovery service. When the fence_kdump agent is invoked,\n"
                      "it will listen for a message from the failed node that acknowledges\n"
                      "that the failed node it executing the kdump crash kernel.\n"
-                     "Note that \\fIfence_kdump\\fP is not a replacement for traditional\n"
-                     "fencing methods. The \\fIfence_kdump\\fP agent can only detect that a\n"
+                     "Note that fence_kdump is not a replacement for traditional\n"
+                     "fencing methods. The fence_kdump agent can only detect that a\n"
                      "node has entered the kdump crash recovery service. This allows the\n"
                      "kdump crash recovery service complete without being preempted by\n"
                      "traditional power fencing methods.\n\n"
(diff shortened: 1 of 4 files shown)


oalbrigt:
These are used for formatting the manpages, so I've made a separate manpage function for the Python agents to be able to output the metadata without formatting, so you're welcome to update it to include these `\f` formatters as well:
https://github.com/ClusterLabs/fence-agents/blob/b6fe4dc4c742dc1d9375db9a1b1527cd418b6ebd/lib/fencing.py.py#L575

OndrejHome:
Thank @oalbrigt , that helps to resolve the formatting in `fence_vmware` using filter below:
~~~
docs["longdesc"] = re.sub("\\\\f[BPIR]|\.P|\.TP|\.br\n", "", docs["longdesc"])
~~~
However the `fence_kdump` is compiled from C instead of using python so that approach won't work there. From the Makefile it looks that we first compile the binary and then extract the manpage from it by calling `-o metadata`. That however means that formatting strings for man page will be always present in `-o metadata` output.

https://github.com/ClusterLabs/fence-agents/blob/b6fe4dc4c742dc1d9375db9a1b1527cd418b6ebd/agents/Makefile.am#L92-L96

Would it be OK to remove these formatting strings just in `fence_kdump` which is compiled from C, but for the `fence_vmware` use the filtering function so the formatting is preserved?

My main motivation here is to make `fence_kdump` output in `pcs stonith describe fence_kdump` more readable. `fence_vmware` was just a random hit for what had same formatting strings.

oalbrigt:
It seems like we dont format the agent name in the other agents, so I guess that's OK to remove. And I guess you could base the parameters on how it's done in the other agents' metadata/manpage from e.g. `man fence_scsi`.

OndrejHome:
I will have a look at this today/tomorrow and post comment here once done. My plan is to rework this  in following way:
- `fence_vmware` should be left without change
- filtering for the `-o metadata` long description will get new values to filter out (man page formatting)
- test case for `fence_vmware` needs to be checked if change also applies there
- `fence_kdump` will get longdescription and parameters changed according to other fence agents, which most probably means that formatting marks for man page will be removed (not replaced with quotes)

I wonder if the tests for `fence_kdump` needs to be somehow changed as when I have forgotten to do that for `fence_vmware` the CI complained. But there was no complain for `fence_kdump`,

oalbrigt:
Just run `make xml-upload` to update the metadata after updating the agent.

OndrejHome:
PR reworked, fence_kdump now don't use the formatting strings for man as the other fence-agents.
I have manually rebuild the `tests/data/metadata/fence_kdump.xml` as the `make xml-upload` nor `xml-check` is doing that for `fence_kdump`.

oalbrigt:
LGTM. Thanks.

PR #283, opened 2019-08-13
fence_compute/fence_evacuate: fix argument parsing for region_name

gpocentek:
The region_name attribute for fence_compute and fence_evacuate is not a boolean
and should expect a string argument.


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #304, opened 2019-10-16
fence_rhevm: added cookie file management

tothf:
We are using fence_rhevm many places but it was annoyingly spamming the RHV manager event logs with a new login every time the agent run status or any other actions. Reuse cookie was handy in case another action happened during the same session.
The change introduce a new option, --cookie-file. It should be full path where the cookie file should be created. By default the name is generated from the IP+username and the file created under /tmp.
This way the agent will store the cookie and use the same session until it's available. In case of manager restart new session will be opened.
In case the cookie file is stored on clustered filesystem the session will be kept in case of cluster node restarts or resource move.

### DIFF
--- a/agents/rhevm/fence_rhevm.py
+++ b/agents/rhevm/fence_rhevm.py
@@ -117,7 +117,12 @@ def send_command(opt, command, method="GET"):
 		conn.setopt(pycurl.HTTPAUTH, pycurl.HTTPAUTH_BASIC)
 		conn.setopt(pycurl.USERPWD, opt["--username"] + ":" + opt["--password"])
 		if "--use-cookies" in opt:
-			conn.setopt(pycurl.COOKIEFILE, "")
+			if "--cookie-file" in opt:
+				cookie_file = opt["--cookie-file"]
+			else:
+				cookie_file = "/tmp/fence_rhevm_" + opt["--ip"] + "_" + opt["--username"] + "_cookie.dat"
+			conn.setopt(pycurl.COOKIEFILE, cookie_file)
+			conn.setopt(pycurl.COOKIEJAR, cookie_file)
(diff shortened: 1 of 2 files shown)


oalbrigt:
Looks good to me, but you also need to add the new parameter this array for it to show up in usage/metadata: https://github.com/ClusterLabs/fence-agents/blob/b885e0f65af626154096a49554e9765e18bfbbd9/agents/rhevm/fence_rhevm.py#L209

After you do that you'll have to run `make xml-upload` to update tests metadata, or our CI will fail.

tothf:
> Looks good to me, but you also need to add the new parameter this array for it to show up in usage/metadata:
> 
> https://github.com/ClusterLabs/fence-agents/blob/b885e0f65af626154096a49554e9765e18bfbbd9/agents/rhevm/fence_rhevm.py#L209
> 
> After you do that you'll have to run `make xml-upload` to update tests metadata, or our CI will fail.

I committed the change, tried to run make xml-upload with agentpycheck.mk but it didn't work. Either "Nothing to be done" or "No rule to make target" if I specify TARGET.
Can you point me to the right direction?
Thanks

oalbrigt:
agentpycheck.mk is used by the build system, so you dont need to use it in relation to that.

Just run the following commands in the root directory of the project:
```
./autogen.sh
./configure
make xml-upload
```

tothf:
> agentpycheck.mk is used by the build system, so you dont need to use it in relation to that.
> 
> Just run the following commands in the root directory of the project:
> 
> ```
> ./autogen.sh
> ./configure
> make xml-upload
> ```

Thanks. Looks better now :)

oalbrigt:
LGTM. Thanks.

PR #317, opened 2020-01-16
modify 'InstanceName' to 'HostName'

bd233:
Should return hostname instead of VM's InstanceName


bd233:
> Can one of the admins verify this patch?

Yes, they can

oalbrigt:
It's supposed to use instance name, not hostname.

It is up to the user if they want to use same name for instance name and hostname or not, and they can use pcmk_host_map if they are different.

PR #321, opened 2020-02-16
configure: fix agent filtering

vvidic:
If some autodetect agent is last in the list it will not be filtered
(no trailing space) and the build will fail.


oalbrigt:
LGTM. Thanks.

PR #323, opened 2020-03-19
fence_aws: Fix fence race, logging improvement and new debug option

gguifelixamz:
Hello,

This PR has the objective to improve overall fence_aws functionality aligned with customer and field feedback.

###  What is Changing?

- **Fix fence Race condition**
Added a function called ```get_self_power_status()``` in which the instance executing the fence action will check its own status before fencing the other node. If the local EC2 instance's status is not "Running" it will not try to fence the other node as the local node has been already fenced. With the addition of this feature ```pcmk_delay_max``` and/or ```pcmk_delay_base``` can now be used and are encouraged to be used with fence_aws.

- **Logging Changes**
Stopped using parent "logging" from lib fencing.py and added its own local logger. The reason for this is to 1) avoid logs from being streamed to _stderr_; and 2) avoid hitting [RHBZ #1549366](https://bugzilla.redhat.com/show_bug.cgi?id=1549366) when `verbose` logging is enabled.
Also, now while using `verbose` option it will create `/var/log/fence_aws_debug.log` as by default on some distributions (Red Hat) rsyslogd isn't configured to stream `*.debug` messages to syslog. Also with `verbose` option enabled boto3 library will no longer get into debug mode to avoid its logging from flooding the log file (see next item for detail).

- **New Boto3 debug option**
Added a new option to enable Boto3 debug (**--boto3_debug**) which by default is set to "off". Once enabled it will create `/var/log/fence_aws_boto3.log` with all boto3 and AWS APIs debugging information. This is important as previously STONITH `verbose` option was being too noisy and capturing unnecessary information.

### DIFF
--- a/agents/aws/fence_aws.py
+++ b/agents/aws/fence_aws.py
@@ -121,7 +121,7 @@ def define_new_opts():
 	all_opt["boto3_debug"] = {
 		"getopt" : "b:",
 		"longopt" : "boto3_debug",
-		"help" : "-b, --boto3_debug=on|off      Boto3 and Botocore library debug logging",
+		"help" : "-b, --boto3_debug=[on|off]      Boto3 and Botocore library debug logging",
 		"shortdesc": "Boto Lib debug",
 		"required": "0",
 		"order": 5
--- a/agents/aws/fence_aws.py
+++ b/agents/aws/fence_aws.py
@@ -121,7 +121,7 @@ def define_new_opts():
 	all_opt["boto3_debug"] = {
 		"getopt" : "b:",
 		"longopt" : "boto3_debug",
-		"help" : "-b, --boto3_debug=[on|off]      Boto3 and Botocore library debug logging",
+		"help" : "-b, --boto3_debug=[option]      Boto3 and Botocore library debug logging",
 		"shortdesc": "Boto Lib debug",
 		"required": "0",
 		"order": 5
--- a/agents/aws/fence_aws.py
+++ b/agents/aws/fence_aws.py
@@ -197,5 +197,4 @@ For instructions see: https://boto3.readthedocs.io/en/latest/guide/quickstart.ht
 	sys.exit(result)
 
 if __name__ == "__main__":
-
 	main()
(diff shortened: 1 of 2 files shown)


gguifelixamz:
Not sure if the Travis failure is an issue or not -- I rather wait for one of the admins (@oalbrigt) to look

oalbrigt:
You need to run "make xml-upload" to update the metadata CI checks against, and commit the updated xml-file.

oalbrigt:
You might want to set more of the info-logging to debug-level, and set the default to WARN/ERR (cant remember what's the usual default), to avoid spamming the logs unneccesarily.

gguifelixamz:
Thanks for the help @oalbrigt it passed now :) :)

agents/aws/fence_aws.py
```diff
@@ -122,4 +191,12 @@ def main():
 	sys.exit(result)
 
 if __name__ == "__main__":
+
```
oalbrigt: I think it'd be better to put this code in top of the main()-function, so it's more logical when looking through the logging code.

gguifelixamz: I tried, but then "logger" won't be a global logger within fence_aws but only to `main()`, and instantiating a new logger at every function call seems like too much, so although the code doesn't look elegant it is the only one I could make it to deliver the results I wanted (as far as logging is concerned).

The main idea is to prevent the usage of the root logger (inherited from "lib/fencing.py") and to have a single-local logger within the fence agent.

oalbrigt: Ah. In that case I think it'd be better to put it after the imports (so people see that they're global easily).

gguifelixamz: Done! Done! 😄

oalbrigt: Can you also remove that empty line?

gguifelixamz: > Can you also remove that empty line?

Empty line removed!

oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

oalbrigt:
Patch for Python 3: https://github.com/ClusterLabs/fence-agents/pull/324

PR #326, opened 2020-04-13
fence_gce: Disable google api cache discovery

arbulu89:
The underlying google api code cache discovery process
uses deprecated code. In order to avoid the usage of that
code and the related warning messages the cache usage is
disabled

### DIFF
--- a/agents/gce/fence_gce.py
+++ b/agents/gce/fence_gce.py
@@ -230,7 +230,8 @@ def main():
 		if tuple(googleapiclient.__version__) < tuple("1.6.0"):
 			import oauth2client.client
 			credentials = oauth2client.client.GoogleCredentials.get_application_default()
-		conn = googleapiclient.discovery.build('compute', 'v1', credentials=credentials)
+		conn = googleapiclient.discovery.build(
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

agents/gce/fence_gce.py
```diff
@@ -230,7 +230,8 @@ def main():
 		if tuple(googleapiclient.__version__) < tuple("1.6.0"):
 			import oauth2client.client
 			credentials = oauth2client.client.GoogleCredentials.get_application_default()
-		conn = googleapiclient.discovery.build('compute', 'v1', credentials=credentials)
+		conn = googleapiclient.discovery.build(
+            'compute', 'v1', credentials=credentials, cache_discovery=False)
```
oalbrigt, line 234: Can you indent this a bit more, so it's clear it's part of the line above?

arbulu89: Sure. For some reason `tabs` are used instead of `spaces` and my editor renders different... (that's the risk of using tabs...).

oalbrigt:
LGTM. Thanks.

PR #336, opened 2020-05-16
fence_openstack: import novaclient and keystone only when needed

vvidic:
Remove build dependency and always allow printing of metadata
similar to fence_compute and fence_evacuate.


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

oalbrigt:
LGTM. Thanks.

PR #346, opened 2020-07-06
fencing: Add verbose_level option and make verbose flags incrementable

nrwahl2:
Currently, `verbose` is a boolean, all-or-nothing option. It would be
useful to be able to set more granular verbosity levels when debugging
an issue.

This patch adds a `verbose_level` option to allow setting higher
verbosities. This option can be set explicitly (e.g.,
`--verbose-level=3`). Alternatively, multiple `verbose` flags (e.g.,
`-vvv`) can be passed on the command line; `verbose_level` is set to
the number of `-v` flags.

If both an explicit `verbose_level` and one or more `-v` options are
set, the `verbose_level` takes precedence and overrides any `-v` flags.
If `verbose_level > 0` and there are no `-v` flags, the `verbose`
option is enabled implicitly.

Later patches will enable individual fence agents to use the new
verbosity levels.

Related to: RHBZ#1853973

### DIFF
--- a/lib/fencing.py.py
+++ b/lib/fencing.py.py
@@ -38,27 +38,40 @@ all_opt = {
 		"help" : "-h, --help                     Display this help and exit",
 		"required" : "0",
 		"shortdesc" : "Display help and exit",
-		"order" : 54},
+		"order" : 55},
 	"version" : {
 		"getopt" : "V",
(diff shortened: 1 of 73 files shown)


oalbrigt:
Tested with Python 2 and 3, and everything works as expected.

PR #348, opened 2020-07-15
Make sure fence-compute depends on novaclient

lmiccini:
fence-compute depends on novaclient to work, so this commit adds
the explicit dependency to make sure the necessary rpm is
installed along the fence-agent.


oalbrigt:
add to whitelist

oalbrigt:
Thanks.

PR #350, opened 2020-07-24
[fence_evacuate] Enable evacuation of instances using private flavors

lmiccini:
This commit extends the flavor.list() api call to fetch private
flavors that could be tagged with the 'evacuable' attribute,
essentially allowing instance-ha to be enabled on a per tenant basis.


lmiccini:
relevant rhbz: https://bugzilla.redhat.com/show_bug.cgi?id=1859932

oalbrigt:
LGTM. Thanks.

PR #351, opened 2020-07-25
fence_lpar: Fix parse error from long command line

nrwahl2:
When Pacemaker executes `fence_lpar` and the HMC command line is greater
than 80 characters, a parse error causes agent failure. This can happen
with a long user name and/or long managed system name. It happens only
when Pacemaker spawns the `fence_lpar` process; it does not happen when
`fence_lpar` is run from the CLI.

A long command line gets a carriage return ('\r') added at the 80
character mark and wraps back to the beginning of the line with no line
feed ('\n'), overwriting the displayed characters. `fence_lpar`'s regex
matches handle this fine when it's run from the command line.

The problem is that when Pacemaker spawns fence_lpar, **for some
reason** there are backspace characters in the buffer when we hit the
'\r' character. This seems to overwrite some of the `conn.before`
string. As a result, the regex doesn't match `conn.before`, and the
agent fails.

This patch works around the `conn.before` weirdness by reading and
discarding the first received line **before** any regex processing.

Resolves: RHBZ#1860544
Resolves: RHBZ#1860545

Signed-off-by: Reid Wahl <nrwahl@protonmail.com>

---

Below is an example of a monitor failure due to this issue:
```
Jul 24 17:20:13 fastvm-rhel-8-0-24 pacemaker-fenced[1607]: warning: fence_lpar[4419] stderr: [ 2020-07-24 17:20:13,074 DEBUG: Sent: lssyscfg -r lpar -m ibm-p9z-20 -F name:state ]
Jul 24 17:20:13 fastvm-rhel-8-0-24 pacemaker-fenced[1607]: warning: fence_lpar[4419] stderr: [  ]
Jul 24 17:20:13 fastvm-rhel-8-0-24 pacemaker-fenced[1607]: warning: fence_lpar[4419] stderr: [ 2020-07-24 17:20:13,238 DEBUG: Received: ' lssyscfg -r lpar -m ibm-p9z-20 -F na\r<h@p9-vhmc1'':~>' ]
Jul 24 17:20:13 fastvm-rhel-8-0-24 pacemaker-fenced[1607]: warning: fence_lpar[4419] stderr: [ 2020-07-24 17:20:13,240 ERROR: Unable to parse output of list command ]
```

Below you can see the difference in `conn.buffer` and `conn.before` when this is run from the CLI vs. from Pacemaker.
```
CLI:
buffer (last 100 chars): ' '
before (last 100 chars): '\np9-node4:Running\r\np9-node2:Running\r\nibm-p9z-20-vios1:Running\r\nsv-hanahmc_prd-aunxhmc2a-z2h@p9-vhmc1'
after: ':~>'

Pacemaker:
buffer (last 100 chars): ' lssyscfg -r lpar -m ibm-p9z-20 -F nam                         \x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08e:state\r\n'
before (last 100 chars): ' lssyscfg -r lpar -m ibm-p9z-20 -F na\r<h@p9-vhmc1'
after: ':~>'
```

### DIFF
--- a/agents/lpar/fence_lpar.py
+++ b/agents/lpar/fence_lpar.py
@@ -19,6 +19,9 @@ from fencing import fail, fail_usage, EC_STATUS_HMC
 def get_power_status(conn, options):
 	if options["--hmc-version"] == "3":
 		conn.send("lssyscfg -r lpar -m " + options["--managed"] + " -n " + options["--plug"] + " -F name,state\n")
+
+		# First line (command) may cause parsing issues if long
+		conn.readline()
 		conn.log_expect(options["--command-prompt"], int(options["--power-timeout"]))
 
 		try:
@@ -29,6 +32,9 @@ def get_power_status(conn, options):
 	elif options["--hmc-version"] in ["4", "IVM"]:
 		conn.send("lssyscfg -r lpar -m "+ options["--managed"] +
 				" --filter 'lpar_names=" + options["--plug"] + "'\n")
+
+		# First line (command) may cause parsing issues if long
+		conn.readline()
 		conn.log_expect(options["--command-prompt"], int(options["--power-timeout"]))
 
 		try:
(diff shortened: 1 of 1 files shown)


nrwahl2:
Full disclosure, the changes in the `hmc-version="3"` and `hmc-version="ivm"` sections are based on an **assumption** that they are prone to the same issue as HMC version >=4 and that the same solution applies. From what I can tell looking at the regex, adding `readline` and modifying the regexes as I have done should be safe even if those versions are not prone to this issue.

PR #352, opened 2020-07-29
fence_ibmz: fix handling of API response when LPAR is not found

ppinatti:
According to its documentation the HMC API should return an
empty array when an LPAR is not found but instead for CPCs in classic
mode it returns a 404 response. This patch changes the
handling of the response to handle the situation accordingly.

Signed-off-by: Paulo de Rezende Pinatti <ppinatti@linux.ibm.com>


oalbrigt:
ok

oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #353, opened 2020-07-30
 fence_lpar: Fix list-status action

nrwahl2:
The `list-status` action prints "UNKNOWN" status for all LPARs when
`--hmc-version` is `"4"` or `"IVM"`.

Here we fix that by mapping the statuses returned by the HMC
(e.g., "Running") to the statuses that the fencing library expects
(e.g., "on").

We also refactor `get_lpar_list` to reduce code duplication.

Resolves: RHBZ#1861926


oalbrigt:
LGTM. Thanks.

PR #360, opened 2020-10-23
Add a fence_crosslink agent

mbaldessari:
As suggested by Andrew we came up with this small agent
to cater for a very specific two-node use-case where a crosscable
connects the two nodes and there is no external PDU fencing
available (aka only the integrated BMC is usable).
With this agent we can make the assumption that if the crosscable
is not available the other node has been powered off.

It is meant to be used with the normal BMC/IPMI fencing as a second
level. See the README.md file for more information.

Co-Authored-By: Damien Ciabrini <dciabrin@redhat.com>

### DIFF
--- a/agents/crosslink/README.md
+++ b/agents/crosslink/README.md
@@ -1,0 +1,44 @@
+# Two node cross-link fence agent
+
+The problem that this fence agents tries to solve is the following:
+
+Given a two-node cluster with a direct crosslink ethernet cable
+between the two nodes (in addition to the normal networking setup), we want to
+be able to maintain quorum on node (A) when node (B) lost power.
+The loss of power on node (B) in this case implies its BMC/IPMI is also
+not available which would be normally used in fencing in this case.
+
+Note: An external PDU would be preferrable and would solve this
+situation more elegantly. The assumption here is that something
+like that won't be available in this environment.
+
+This works by creating a stonith level composed of a BMC/IPMI
+fencing at level 1 and then the fence_crosslink agent at level 2.
+ 
+In case node (A) has lost power, then node (B) will do the following:
+1. Try to fence node (B) via IPMI, which will fail since the node has no
+power and the BMC is unavailable
+2. Check via fence_crosslink the cross-cable interconnect. If the cross cable
+IP is not reachable, then we know for "sure" (this is a potentially broad
+assumption) that the node is really down and fence_crosslink tells pacemaker
+that the fencing was successful, so pacemaker can work with that new
+information.
+ 
+Here are some example configuration commands:
+~~~
+pcs stonith create crosslink-controller-1 fence_crosslink crosscableip=1.1.1.2 pcmk_host_list=controller-1 pcmk_reboot_action=off
+pcs stonith create crosslink-controller-0 fence_crosslink crosscableip=1.1.1.1 pcmk_host_list=controller-0 pcmk_reboot_action=off
+# We make sure the stonith resource do not run on the same node as the fencing target
+pcs constraint location crosslink-controller-1 avoids controller-1
+pcs constraint location crosslink-controller-0 avoids controller-0
+pcs stonith level add 2 controller-0 crosslink-controller-0
+pcs stonith level add 2 controller-1 crosslink-controller-1
+~~~
+
+Testing done:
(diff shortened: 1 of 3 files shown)


fabbione:
Copy paste from IRC conversation with @mbaldessari 

16:30 < fabbione> bandini: i keep thinking about that agent and no.. it´s just wrong with ping
16:30 < fabbione> bandini: you need to use ethtool to detect the ethernet electrical status
16:31 < fabbione> say the host is half dead, it might not be answering to ping
16:31 < fabbione> but not fully dead
16:31 < fabbione> somebody run the wrong iptables or firewall command and ping stops responding
16:31 < fabbione> so ping itself is not reliable enough
16:31 < fabbione> ethtool instead will just talk to the card to get electric status of the wire (connected or not)
16:31 < bandini> it is not trying to save the world. it covers a case which might be reasonable or not depending on expectations
16:32 < fabbione> it doesn´t require any response from the other node at all (ICMP_ECHO_REPLY)
16:32 < fabbione> right, that I fully understand, but it can be more reliable
16:32 < fabbione> i like the concept to use the network link as detector
16:32 < fabbione> but then we can do a better detection
16:34 < fabbione> bandini: let me add this to the PR so the conversation is recorded for e-ddie and damien

mbaldessari:
I guess we'll discuss it at more length on Monday. I don't disagree that using ethtool removes a layer and could potentially be preferable but it's not such a clear-cut simple decision:
- I cannot use this fence agent on two VMs running on two node-hypervisors
- ethtool does not have a clean API to do that, so I am left with a) either running 'ethtool | grep "Link detected:" and pray none ever changes that string or b) use python-ethtool or the likes and add a new dependency
- testing in VMs becomes a bit cumbersome? (We'll have to down the IP link on the VM side of the cable as downing the link from the hypervisor (the other tap side) does not change the status of the link in the VM. Maybe there are ways around this though)

As said, I am not against using ethtool, but I think the ICMP approach has a certain simplicity and flexibility to it that should not be discarded too quickly. As to the VM over two nodes thing, we genuinely don't know if it will be a must requirement for us or not just yet (it's a long story). ttyt on Monday

fabbione:
On 10/23/2020 10:52 PM, Michele Baldessari wrote:
> I guess we'll discuss it at more length on Monday. I don't disagree that 
> using ethtool removes a layer and could potentially be preferable but 
> it's not such a clear-cut simple decision:

Yes I understand both have pros and cons.

> 
>   * I cannot use this fence agent on two VMs running on two node-hypervisors

The way I understood this agent is that it is for BM only. Adding more 
layers to an ICMP test could make things even worst since we are now 
adding 2 more nodes that could drop packets around in case of 
malfunctioning and real fencing is not available.

Is the HV dropping packets / not responding or is it the VM or is it the 
switch between the HV or the firewalls between machines? etc. etc.

>   * ethtool does not have a clean API to do that, so I am left with a)
>     either running 'ethtool | grep "Link detected:" and pray none ever
>     changes that string or b) use python-ethtool or the likes and add a
>     new dependency

I don´t see adding a dependency as an issue to be honest. Since all 
agents are shipped in their own package, we are not pulling in the world 
for this one.

>   * testing in VMs becomes a bit cumbersome? (We'll have to down the IP
>     link on the VM side of the cable as downing the link from the
>     hypervisor (the other tap side) does not change the status of the
>     link in the VM. Maybe there are ways around this though)

Testing is another "problem". Yes it requires some more mingling with 
the test suite, but we do already a lot of black magic around to 
simulate failures.. one more will just increase our wizard points a bit :)

> 
> As said, I am not against using ethtool, but I think the ICMP approach 
> has a certain simplicity and flexibility to it that should not be 
> discarded too quickly. As to the VM over two nodes thing, we genuinely 
> don't know if it will be a must requirement for us or not just yet (it's 
> a long story). ttyt on Monday

Maybe we should allow usage of both methods. I am not saying using ICMP 
should be discarded, but we still need to try to be as reliable as 
possible and I see ICMP a lot more fragile in general.

mbaldessari:
Thanks for your time today, Fabio. So we'll add the ethtool support in a separate PR in the near future so an operator can choose both mechanisms and we'll favor the ethtool variant on BM setups. (Scream if I am missing something ;)

oalbrigt:
LGTM. Thanks.

PR #372, opened 2020-12-07
Adds service account authentication to GCE fence agent

megelatim:
Adds the ability to use a service acount for authentiation using --serviceaccount.
Adds additional debug information.

### DIFF
--- a/agents/gce/fence_gce.py
+++ b/agents/gce/fence_gce.py
@@ -1,5 +1,13 @@
 #!@PYTHON@ -tt
 
+#
+# Requires the googleapiclient and oauth2client
+# RHEL 7.x: google-api-python-client==1.6.7 python-gflags==2.0 pyasn1==0.4.8 rsa==3.4.2
+# RHEL 8.x: nothing additional needed
+# SLES 12.x: python-google-api-python-client python-oauth2client python-oauth2client-gce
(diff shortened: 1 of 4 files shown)


wenningerk:
test this please

oalbrigt:
We need to fall back to the old solution when running with old libraries for backwards compatibility.

megelatim:
> test this please

Tested - had to add oauth2client to the travis pip installs

> We need to fall back to the old solution when running with old libraries for backwards compatibility.

This was tested on python 2.7 and python 3 (RHEL 7 & 8 & SLES 12 & 15)

oalbrigt:
ok to test

agents/gce/fence_gce.py
```diff
@@ -20,6 +28,8 @@
 
 import googleapiclient.discovery
 from fencing import fail_usage, run_delay, all_opt, atexit_handler, check_input, process_input, show_docs, fence_action
+from oauth2client.client import GoogleCredentials
```
oalbrigt, line 31: Can do do these in a try/except: pass ?

That should make it able to be built (e.g. manpage generation) even on FreeBSD, where we dont have oauth2client currently installed.

megelatim: Added the try/except:pass to the oauth2client.* imports

oalbrigt:
LGTM. Thanks.

PR #379, opened 2021-01-25
fence_vbox: do not flood host account (history) with vboxmanage calls

BlackZork:
This change disables bash history when checking vm status on host.
Useful mainly in devel installations, where fence_vbox.py fills user .bash_history with:

VBoxManage list vms
VBoxManage list runningvms

I think that there is no reason to log those commands in production environment, hence this pull request.

Tested on Centos7.8 and Virtualbox 6.1.16

### DIFF
--- a/agents/vbox/fence_vbox.py
+++ b/agents/vbox/fence_vbox.py
@@ -94,7 +94,7 @@ def main():
     all_opt["secure"]["default"] = "1"
 
     all_opt["cmd_prompt"]["default"] = [r"\[EXPECT\]#"]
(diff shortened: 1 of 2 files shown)


oalbrigt:
Can you prefix the commit with "fence_vbox: " ? You can do it by running `git commit --amend` followed by `git push -f`.

oalbrigt:
You have to run `make xml-upload` and add the updated metadata to avoid CI failing.

oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #382, opened 2021-02-08
fence_gce: Adds cloud-platform scope for bare metal API and optional proxy flags

kj1724:
This change adds the necessary scope for using the fence agent on bare metal. Along with that we have added proxy flags in case the customer is using a proxy for internet access instead of a NAT gateway

### DIFF
--- a/agents/gce/fence_gce.py
+++ b/agents/gce/fence_gce.py
@@ -15,7 +15,6 @@ import json
 import re
 import os
 import socket
-import socks
 import sys
 import time
 
@@ -34,6 +33,7 @@ try:
   import googleapiclient.discovery
   from oauth2client.client import GoogleCredentials
   from oauth2client.service_account import ServiceAccountCredentials
+  import socks
 except:
   pass
 
--- a/agents/gce/fence_gce.py
+++ b/agents/gce/fence_gce.py
@@ -9,13 +9,16 @@
 #
 
 import atexit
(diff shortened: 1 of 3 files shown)


agents/gce/fence_gce.py
```diff
@@ -9,13 +9,16 @@
 #
 
 import atexit
+import httplib2
 import logging
 import json
 import re
 import os
 import socket
+import socks
```
oalbrigt, line 18: You can move this down to the `try/except: pass` below, and you should be able to remove PySocks from `.travis.yml`.

oalbrigt (changes requested):
Can you prefix the commits with `fence_gce: `?

agents/gce/fence_gce.py
kj1724: I moved the import and updated the travis.yml file

oalbrigt:
Will the bare metal scope work for every type of VM/bare metal, or should we make a parameter to be able to change it?

oalbrigt:
ok to test

oalbrigt (approved):

kj1724:
This bare metal scope work will allow the fence_gce agent to use the reset API for Google cloud VM/bare metal instances. We shouldn't need an extra parameter for now.

Thanks for reviewing!

oalbrigt:
Thanks.

PR #383, opened 2021-02-09
fence_gce: update module reqs for SLES 15

stefanotorresi:
We just found out `python3-oauth2client-gce` is required too, in SLES 15.


oalbrigt:
Thanks.

PR #389, opened 2021-03-11
Default method moved back to powercycle

megelatim:
Moving back to powercycle as the default method for GCE.  This is the recommended approach from Google.  Using on/off as the default will have unintended consequences on google cloud platforms.

### DIFF
--- a/agents/gce/fence_gce.py
+++ b/agents/gce/fence_gce.py
@@ -358,6 +358,8 @@ def main():
 	define_new_opts()
 
 	all_opt["power_timeout"]["default"] = "60"
+	all_opt["method"]["default"] = "cycle"
+	all_opt["method"]["help"] = "-m, --method=[method]          Method to fence (onoff|cycle) (Default: cycle)"
 
 	options = check_input(device_opt, process_input(device_opt))
 
--- a/tests/data/metadata/fence_gce.xml
+++ b/tests/data/metadata/fence_gce.xml
@@ -12,7 +12,7 @@ For instructions see: https://cloud.google.com/compute/docs/tutorials/python-gui
 	</parameter>
 	<parameter name="method" unique="0" required="0">
 		<getopt mixed="-m, --method=[method]" />
-		<content type="select" default="onoff"  >
+		<content type="select" default="cycle"  >
 			<option value="onoff" />
 			<option value="cycle" />
 		</content>


oalbrigt:
From https://bugzilla.redhat.com/show_bug.cgi?id=1906978:
"The problem is that the Google Cloud API doesn't declare the `reset` call to be a success until long after the node has already been powered back on. This means the node may rejoin the cluster before the fence action is declared a success."

megelatim:
> From https://bugzilla.redhat.com/show_bug.cgi?id=1906978:
> "The problem is that the Google Cloud API doesn't declare the `reset` call to be a success until long after the node has already been powered back on. This means the node may rejoin the cluster before the fence action is declared a success."

This has been changed on the google API side to shorten the time between requesting 'reset' and having the API return that the operation is complete.  There can still be some time difference, but in most cases it will be much less than what was observed when this change was introduced.  Google will still recommend a startup delay for pacemaker.

Also - the on/off functionality does not operate like a reset.  The API call to off is not a hard power off, this call waits for a clean shutdown from the host.  If the host is hung, the off call will never return.  The only call that produces a hard power off is the reset call (powercycle).  Google also does not offer on/off API calls for the bare metal solution, only the reset api is available.

megelatim:
Can this pull request be looked at again?

oalbrigt:
ok to test

nrwahl2:
FWIW, it looks like a startup delay is no longer needed at all for the common case, although it adds an extra safety net in case of unforeseen issues.

Fencing was initated at 23:17:47 and completed at 23:17:59.
```
Mar 24 23:17:47 nwahl-rhel8-node1 pacemaker-fenced[1322]: notice: gce_fence2 is eligible to fence (reboot) node2 (aka. 'nwahl-rhel8-node2'): static-list
Mar 24 23:17:59 nwahl-rhel8-node1 pacemaker-fenced[1322]: notice: Operation 'reboot' [1840] (call 4 from pacemaker-controld.1326) for host 'node2' with device 'gce_fence2' returned: 0 (OK)
Mar 24 23:17:59 nwahl-rhel8-node1 pacemaker-fenced[1322]: notice: Operation 'reboot' targeting node2 on node1 for pacemaker-controld.1326@node1.e7c02e53: OK
Mar 24 23:17:59 nwahl-rhel8-node1 pacemaker-controld[1326]: notice: Stonith operation 4/1:3:0:739869d8-d4f7-4dad-9be6-2411d9d6dd3c: OK (0)
Mar 24 23:17:59 nwahl-rhel8-node1 pacemaker-controld[1326]: notice: Peer node2 was terminated (reboot) by node1 on behalf of pacemaker-controld.1326: OK
```

The fenced node booted up at 23:18:36.
```
Mar 24 23:18:36 nwahl-rhel8-node2 kernel: Linux version 4.18.0-240.1.1.el8_3.x86_64 (mockbuild@x86-vm-09.build.eng.bos.redhat.com) (gcc version 8.3.1 20191121 (Red Hat 8.3.1-5) (GCC)) #1 SMP Fri Oct 16 13:36:46 EDT 2020
```

nrwahl2:
I'm now noticing the following from Frank Legler in one of the related RHBZs:

"The Cloud API
team is currently working to make the DONE status of the reset operation get
closer to the actual start of the targeted GCE instance - but we will always
have the possibility of the machine being up before the operation reports DONE.
This may be below 10 seconds in some cases, but we cannot guarantee it."

So I suppose that's a "no" to my question of "can we be confident that the reset operation will be marked complete before the reset node boots?" :)

PR #410, opened 2021-05-13
fence_azure_arm: corrections to support Azure SDK >= 15

cleberps:
We have found that fence_azure_arm do not work after Azure SDK >= 15.
We initiated this PR to report the problem and part of the solutions because there are other points that need attention.
Below are them:
1) Back support to older versions of Azure SDK < 15. Code must be improved to detect the version that is running and select functions accordingly.
2) Other functions from agent were not tested yet, but could suffer from same changes in SDK.
Contributions and comments are welcome.

### DIFF
--- a/agents/azure_arm/fence_azure_arm.py
+++ b/agents/azure_arm/fence_azure_arm.py
@@ -115,7 +115,7 @@ def set_power_status(clients, options):
 
         if (options["--action"]=="off"):
             logging.info("Poweroff " + vmName + " in resource group " + rgName)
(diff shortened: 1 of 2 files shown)


diegoakechi:
@cleberps  thanks for initiating this. About backward compatibility, I believe this section of the documentation points some directions on how to do it

https://docs.microsoft.com/en-us/azure/developer/python/azure-sdk-authenticate#credential-object-has-no-attribute-signed_session

oalbrigt:
Closing in favour of reference issue w/backwards compatibility.

PR #415, opened 2021-06-22
fence_azure_arm: corrections to support Azure SDK >= 15 - including backward compatibility

yeoldegrove:
This is based on the work in #410.
It adds some code to check for the possibility to import some modules and access some attributes.
It tries to use the newer module `azure.identity / azure.core` first and falls back to `azure.common` if this fails.

This is not the approach mentioned e.g. here https://docs.microsoft.com/en-us/azure/developer/python/azure-sdk-authenticate#other-authentication-methods / https://gist.github.com/lmazuel/cc683d82ea1d7b40208de7c9fc8de59d or here https://github.com/jongio/azidext/blob/master/python/azure_identity_credential_adapter.py.
The obove approaches did not work out as they import again `azure.core.pipeline` which is only available with newer API versions.

My test systems which run this code flawlessly are:

```
# SUSE Linux Enterprise Server 12 SP3
# python2
sles12sp3:~ # rpm -q python-azure-mgmt python-azure-mgmt-compute python-azure-mgmt-resource
python-azure-mgmt-4.0.0-2.10.1.noarch
python-azure-mgmt-compute-4.6.2-2.6.3.noarch
python-azure-mgmt-resource-2.0.0-2.6.2.noarch

# SUSE Linux Enterprise Server 15 SP2
# python3
sles15sp2:~ # rpm -q python3-azure-mgmt-core python3-azure-mgmt-compute python3-azure-mgmt-resource
python3-azure-mgmt-core-1.2.2-3.3.1.noarch
python3-azure-mgmt-compute-17.0.0-6.7.1.noarch
python3-azure-mgmt-resource-10.2.0-6.4.1.noarch
```

### DIFF
--- a/agents/azure_arm/fence_azure_arm.py
+++ b/agents/azure_arm/fence_azure_arm.py
@@ -120,7 +120,7 @@ def set_power_status(clients, options):
                 compute_client.virtual_machines.begin_power_off(rgName, vmName, skip_shutdown=True)
             except AttributeError:
                 # use older API verson if it fails
-                logging.info("Poweroff " + vmName + " did not work via 'virtual_machines.begin_power_off. Trying virtual_machines.power_off'.")
+                logging.debug("Poweroff " + vmName + " did not work via 'virtual_machines.begin_power_off. Trying virtual_machines.power_off'.")
                 compute_client.virtual_machines.power_off(rgName, vmName, skip_shutdown=True)
         elif (options["--action"]=="on"):
             logging.info("Starting " + vmName + " in resource group " + rgName)
@@ -129,7 +129,7 @@ def set_power_status(clients, options):
                 compute_client.virtual_machines.begin_start(rgName, vmName)
             except AttributeError:
                 # use older API verson if it fails
-                logging.info("Starting " + vmName + " did not work via 'virtual_machines.begin_start. Trying virtual_machines.start'.")
+                logging.debug("Starting " + vmName + " did not work via 'virtual_machines.begin_start. Trying virtual_machines.start'.")
                 compute_client.virtual_machines.start(rgName, vmName)
 
 
--- a/agents/azure_arm/fence_azure_arm.py
+++ b/agents/azure_arm/fence_azure_arm.py
@@ -115,10 +115,22 @@ def set_power_status(clients, options):
 
         if (options["--action"]=="off"):
             logging.info("Poweroff " + vmName + " in resource group " + rgName)
-            compute_client.virtual_machines.begin_power_off(rgName, vmName, skip_shutdown=True)
+            # use older API version first
+            try:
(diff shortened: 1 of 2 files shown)


agents/azure_arm/fence_azure_arm.py
```diff
@@ -115,10 +115,22 @@ def set_power_status(clients, options):
...
+            try:
+                compute_client.virtual_machines.power_off(rgName, vmName, skip_shutdown=True)
+            # try new API version if it fails
+            except AttributeError:
+                logging.info("Poweroff " + vmName + " did not work via 'virtual_machines.power_off. Trying virtual_machines.begin_power_off'.")
+                compute_client.virtual_machines.begin_power_off(rgName, vmName, skip_shutdown=True)
         elif (options["--action"]=="on"):
             logging.info("Starting " + vmName + " in resource group " + rgName)
-            compute_client.virtual_machines.start(rgName, vmName)
+            # use older API version first
```
oalbrigt, line 127: I think we should try new libraries first in the agent too, and fallback to old like in the library, as that'll cause less failed try's in the future.

yeoldegrove: :+1: fixed in 46c0872

agents/azure_arm/fence_azure_arm.py
```diff
@@ -115,10 +115,22 @@ def set_power_status(clients, options):
...
+                compute_client.virtual_machines.power_off(rgName, vmName, skip_shutdown=True)
         elif (options["--action"]=="on"):
             logging.info("Starting " + vmName + " in resource group " + rgName)
-            compute_client.virtual_machines.start(rgName, vmName)
+            try:
+                # try new API version first
+                compute_client.virtual_machines.begin_start(rgName, vmName)
+            except AttributeError:
+                # use older API verson if it fails
+                logging.info("Starting " + vmName + " did not work via 'virtual_machines.begin_start. Trying virtual_machines.start'.")
```
oalbrigt, line 132: These should be debug level, so it doesnt spam the logs unless you set HA_debug=1 or similar.

yeoldegrove: fixed in a0f5193c323d04beb64defca4d6b4bca3631c0b5

oalbrigt:
ok to test

diegoakechi (approved):
LGTM @yeoldegrove!

@oalbrigt  Could you squash this when merging, please? It can make easier backportings and maintenance.  Thanks!

oalbrigt:
> LGTM @yeoldegrove!
> 
> @oalbrigt Could you squash this when merging, please? It can make easier backportings and maintenance. Thanks!
Will do.

oalbrigt:
LGTM. Thanks.

PR #420, opened 2021-07-01
Mid: fence_sbd: A warning message is output when disable-timeout is enabled.

HideoYamauchi:
Hi All,

If disable-timeout is enabled, power_timeout will be processed with unlimited (0).
Since it is smaller than sbd_msg_timeout, a warning log is always output.

If disable-timeout is enabled, no comparison with sbd_msg_timeout is required.

Best Regards,
Hideo Yamauchi.

### DIFF
--- a/agents/sbd/fence_sbd.py
+++ b/agents/sbd/fence_sbd.py
@@ -394,13 +394,14 @@ which can be used in environments where sbd can be used (shared storage)."
         if DEVICE_INIT != return_code:
             exit(return_code)
 
-    # we check against the defined timeouts. If the pacemaker timeout is smaller
-    # then that defined within sbd we should report this.
-    power_timeout = int(options["--power-timeout"])
-    sbd_msg_timeout = get_msg_timeout(options)
-    if power_timeout <= sbd_msg_timeout:
-        logging.warn("power timeout needs to be \
-                greater then sbd message timeout")
+    if options.get("--disable-timeout", "0").lower() in ["0", "no", "off", "false"]:
+        # we check against the defined timeouts. If the pacemaker timeout is smaller
+        # then that defined within sbd we should report this.
+        power_timeout = int(options["--power-timeout"])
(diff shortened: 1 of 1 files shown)


agents/sbd/fence_sbd.py
```diff
@@ -394,13 +394,14 @@ def main():
...
 
-    # we check against the defined timeouts. If the pacemaker timeout is smaller
-    # then that defined within sbd we should report this.
-    power_timeout = int(options["--power-timeout"])
-    sbd_msg_timeout = get_msg_timeout(options)
-    if power_timeout <= sbd_msg_timeout:
-        logging.warn("power timeout needs to be \
-                greater then sbd message timeout")
+    if options.get("--disable-timeout", "0").lower() in ["0", "no", "off", "false"]:
+        # we check against the defined timeouts. If the pacemaker timeout is smaller
```
wenningerk, line 398: Is there a reason why we can't simply check power_timeout > 0 before issuing a warning?
This way it would suppress the warning as well if power_timeout got disabled directly and not via dissable-timeout.

HideoYamauchi: Hi Klaus,

I think it's okay to just check that power_timeout is greater than 0.

Do you think that is better?
Checking will be simpler.

Best Regards,
Hideo Yamauchi.

oalbrigt: +1 to check that it's greater than 0.

Just update old line 401 to `if 0  < power_timeout <= sbd_msg_timeout:`.

HideoYamauchi: Hi Oyvind,

Thanks for your comment.

> +1 to check that it's greater than 0.
> 
> Just update old line 401 to `if 0 < power_timeout <= sbd_msg_timeout:`.

Okay!

Thanks,
Hideo Yamauchi.

PR #424, opened 2021-07-19
fence_pve: Replace `nodename` with `pmx-node` in fence_pve.py (matching original intent)

broadstack-au:
refactor to suit original intent of internal `nodename` logic.

Change;
 * `pmx-node` option is used to specify the proxmox cluster node controlling the vm
 * `nodename` option indicates the "node to fence"

`fence_pve` relies on `plug` for node actions and never `nodename` - this is unchanged

Current behaviour;
If nodename is not provided to the fence agent, autodiscovery will fail during fencing activities.
 * The "watch/list" behaviours would execute correctly, (due to no nodename provided for those actions).
 * Status and all node-focused actions would fail due to nodename being the cluster node, not the proxmox node. ​

New behaviour;
 * ​If `pmx-node` is provided to the fence agent, it will use that node
 * ​if `pmx-node` is not provided to the fence agent, it will discover and set it correctly

This will break existing implementations that specify the node name - I'm open to suggestions for how to deal with that

### DIFF
--- a/agents/pve/fence_pve.py
+++ b/agents/pve/fence_pve.py
@@ -11,7 +11,7 @@ import io
 import atexit
 import logging
 sys.path.append("@FENCEAGENTSLIBDIR@")
-from fencing import fail, EC_LOGIN_DENIED, atexit_handler, all_opt, check_input, process_input, show_docs, fence_action, run_delay
+from fencing import fail, fail_usage, EC_LOGIN_DENIED, atexit_handler, all_opt, check_input, process_input, show_docs, fence_action, run_delay
 
 if sys.version_info[0] > 2: import urllib.parse as urllib
 else: import urllib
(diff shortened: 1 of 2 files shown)


oalbrigt:
We have to keep nodename, with comment "Replaced by " to avoid breaking setups for current users:
https://github.com/ClusterLabs/fence-agents/blob/master/agents/mpath/fence_mpath.py#L232-L239
https://github.com/ClusterLabs/fence-agents/blob/master/agents/mpath/fence_mpath.py#L287-L294

You can run `make xml-upload` to update the the metadata files.

Commits should be prefixed by the agent name.

agents/pve/fence_pve.py
```diff
@@ -140,14 +140,23 @@ def main():
...
+		"getopt" : ":",
 		"longopt" : "nodename",
-		"help" : "-N, --nodename                 "
+		"help" : "--nodename                 "
+			"Replaced by -N, --pmx-node",
+		"required" : "0",
+		"shortdesc" : "Replaced by -N/--pmx-node",
+		"order": 2
+	}
+	all_opt["pmx_node"] = {
```
oalbrigt, line 151: I think it's better to put the new/sensibly named parameter first, so users see it before the backwards-comptible one.

broadstack-au:
Okay, I'm as happy with this as I think I'll be,

`nodename` is getting injected by fenced as "the node to fence" in some actions (because it should - that's what `nodename` is for), but it's being (mis)used in fence_pve to specify the specific proxmox cluster node that the vm is running on. This means It can't be cleanly replaced without inheriting the old runtime behaviour.

Given cluster node discovery is broken currently (fencing _will_ fail if the nodename parameter is not specified in the fencing config), I've introduced `pve-node` as a replacement input parameter for `nodename` and `pve-node-auto` to force autodiscovery.

Those using the old `nodename` param (incl shortform `N`) won't see a behaviour change.
Updated/new implementations can choose to use `pve-node` or `pve-node-auto`.

Ready for re-review, thanks :)

oalbrigt:
ok to test

oalbrigt:
You need to run `make xml-upload` and attach the updated metadata to the commit, so it can pass the CI tests.

broadstack-au:
> You need to run `make xml-upload` and attach the updated metadata to the commit, so it can pass the CI tests.

sorry about that - forgot to rerun after changing ordering. fixed.

oalbrigt:
No worries. Now we just need to fix the CI side of things :smiley:

oalbrigt:
retest this please

oalbrigt:
LGTM. Thanks.

PR #425, opened 2021-07-19
fence_raritan_px3: new fence agent

watologo1:
Tested on model PX3-5466V with firmware version: 3.6.0.5

### DIFF
--- a/agents/raritan_px3/fence_raritan_px3.py
+++ b/agents/raritan_px3/fence_raritan_px3.py
@@ -1,0 +1,195 @@
+#!@PYTHON@ -tt
+
+import logging
+import sys
+import atexit
+sys.path.append("@FENCEAGENTSLIBDIR@")
+from fencing import *
+from fencing import EC_STATUS
+
+"""
+Raritan PX3 family is totaly different the PX family (PX2 seem to be
+compatible with PX3 and seem to share the same BIOS, so this fence should
+work with PX2 as well).
+It has another command line prompt and totally other commands
+and output.
+
+It follows the concept of separating outlets and outletgroups (if created).
+You can reach outlets via a fixed, not changeble "plug" number
+(from 1-20 on my device).
+Additionally one can, but does not need to assign names to each plug.
+
+Plugs/outlets can be combined to outletgroups.
+There can be zero to N (N = No. outlets) outletgroups.
+
+While it's possible to create outletgroups with one plug, this does not
+make sense and might slow things down.
+
+--plug=X paramter can be:
+1. X == outlet No
+2. X == outlet Name (if one got assigned)
+3. X == outlet group Name
+
+-> One cannot reach a group by number
+-> Groups need an extra call (first single outlet devices are
+   searched for given No/Name, then OutletGroups
+"""
+
+
+class FenceRaritanPX3:
+	outlets={}
+	# Plug id of outlet
+	plug=None
+	outletgroups={}
+	# Group name if outlet plug id/name have not been found
+	group_name=None
+
+def px3_get_outlet_group(conn, options):
+
+	conn.send_eol("show outletgroups")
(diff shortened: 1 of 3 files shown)


watologo1:
While I tested these features to work:
- on/off/status/list/list-status
- single outlets/outletgroups

I run into following issues:
- I had 2 systems, one had rather long network latencies and it costed me quite some time, that it's a simple timeout I was running into while it worked the some time before. The other system was much faster, still I increased the default timeout as it's really nasty to struggle with such an issue.
- Doing on/off switching, one gets asked whether you really want to do it. I searched the config, but could not find anything to switch that off. Sending 'y' works and also the next prompt is correctly "expected".  Means I see the XXXXX debug message I left in there immediately:
```
        print("YYYYY")
        conn.log_expect(options["--command-prompt"], int(options["--shell-timeout"]))
        print("XXXXXXXX")
```
Then I it always hangs until the connection times out.
I tried to additionally expect a ".*" to get through without success.
Some help/hint would be appreciated what/why it hangs there.

Thanks.

agents/raritan_px3/fence_raritan_px3.py
```diff
@@ -0,0 +1,195 @@
...
+		# fetch state of previously parsed outlet from next line/iter
+		if len(split_line) == 3 and split_line[0] == "Power" and split_line[1] == "state:":
+			outlet_state = split_line[2]
+			FenceRaritanPX3.outlets[outlet_no] = [outlet_name, outlet_state]
+	logging.debug("Outlets found:\n%s", FenceRaritanPX3.outlets)
+	return FenceRaritanPX3.outlets
+
+def get_power_status(conn, options):
+
+	if FenceRaritanPX3.plug:
```
oalbrigt, line 107: What status's does it report? on/off/unknown is what the fencing library expects.

watologo1: Yes, on/off

Originally On/Off as shown in the comment:
                Power state: On
(credentials removed):

I save all in an outlet dict, if necessary a second call also
storing outletgroup data is done:

 ./fence_raritan_px3 ... --plug=20  --action=list-status
1,,On
2,,On
3,,On
4,,On
5,,On
6,,On
7,,On
8,Testung,On
9,,On
10,,On
11,,On
12,,On
13,,On
14,mustang1,On
15,mustang2,On
16,mustang4,On
17,mustang3,On
18,,On
19,,On
20,,On

Some plugs/outlets do have names, some can only be accessed via ID.
Raw dict:
{'1': ['', 'On'], '2': ['', 'On'], '3': ['', 'On'], '4': ['', 'On'], '5': ['', 'On'], '6': ['', 'On'], '7': ['', 'On'], '8': ['Testung', 'On'],
'9': ['', 'On'], '10': ['', 'On'], '11': ['', 'On'], '12': ['', 'On'], '13': ['', 'On'], '14': ['mustang1', 'On'],
'15': ['mustang2', 'On'], '16': ['mustang4', 'On'], '17': ['mustang3', 'On'], '18': ['', 'On'], '19': ['', 'On'], '20': ['', 'On']}

I could (should?) already convert the uppercase to lower?
I do this later:
        if FenceRaritanPX3.plug:
                return FenceRaritanPX3.outlets[str(FenceRaritanPX3.plug)][1].lower()
        elif FenceRaritanPX3.group_name:
                return FenceRaritanPX3.outletgroups[FenceRaritanPX3.group_name][1].lower()

So on/off can be returned.

oalbrigt: Sounds good.

Yeah. That makes sense, to make it the same as the other agents, which usually maps the statuses.

Example:
https://github.com/ClusterLabs/fence-agents/blob/master/agents/vmware_rest/fence_vmware_rest.py#L14

PR #435, opened 2021-09-23
README.md: replace freenode with Libera

setharnold:
The IRC channel is on Libera now.


oalbrigt:
I guess we should just point to the IRC channel page on the the clusterlabs wiki like we do in the pacemaker README:
https://github.com/ClusterLabs/pacemaker/blob/master/README.markdown#where-can-i-find-more-information-about-pacemaker

oalbrigt:
Thanks. Merged with link to wiki IRC page in https://github.com/ClusterLabs/fence-agents/pull/561.

PR #439, opened 2021-09-28
azure_fence: fix support for sovereign clouds and MSI for new versions of azure libraries

MSSedusch:
fix support for Azure Sovereign clouds like China or USGov and MSI support


oalbrigt:
We still need to fallback to the old code for distros that havent go the latest versions of the libs.

MSSedusch:
@oalbrigt can you recheck? It should be backward compatible now

oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

Issue #446, opened 2021-11-01
fence_vmware_soap has no connection timeout, results in filling up tmp directory


thanasis2028:
The resource agent `fence_vmware_soap` has no connection timeout when creating a suds `Client` and does not respect any of the `--power-timeout` `--shell-timeout` `--login-timeout` options.

This is important because when running under pacemaker and trying to contact a server that doesn't respond, the script will run until pacemaker forcibly kills it, which leaks the created python `tmpdir` (and will eventually fill up the disk).

It can be easily observed by setting up fence_vmware_soap in pacemaker and then blocking the server in iptables:
```
pcs stonith create stonith_vmware fence_vmware_soap ipaddr="${esxi_host_ip}" ....
iptables -A INPUT -j DROP -s "${esxi_host_ip}"
```
And the `/tmp/` directory will slowly start filling up.

oalbrigt:
retest this please

PR #447, opened 2021-11-01
fence_vmware_soap: Use --login-timeout option

thanasis2028:
Fixes issue #446.

### DIFF
--- a/agents/vmware_soap/fence_vmware_soap.py
+++ b/agents/vmware_soap/fence_vmware_soap.py
@@ -57,7 +57,7 @@ def soap_login(options):
 
 	try:
 		headers = {"Content-Type" : "text/xml;charset=UTF-8", "SOAPAction" : "vim25"}
-		login_timeout = 60 if "--disable-timeout" in options and options["--disable-timeout"] != "false" else int(options["--login-timeout"])
+		login_timeout = int(options["--login-timeout"]) or 60
 		conn = Client(url + "/vimService.wsdl", location=url, transport=RequestsTransport(verify=verify), headers=headers, timeout=login_timeout)
 
 		mo_ServiceInstance = Property('ServiceInstance')
--- a/agents/vmware_soap/fence_vmware_soap.py
+++ b/agents/vmware_soap/fence_vmware_soap.py
@@ -57,7 +57,7 @@ def soap_login(options):
 
 	try:
 		headers = {"Content-Type" : "text/xml;charset=UTF-8", "SOAPAction" : "vim25"}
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

oalbrigt:
retest this please

agents/vmware_soap/fence_vmware_soap.py
```diff
@@ -57,7 +57,8 @@ def soap_login(options):
 
 	try:
 		headers = {"Content-Type" : "text/xml;charset=UTF-8", "SOAPAction" : "vim25"}
-		conn = Client(url + "/vimService.wsdl", location=url, transport=RequestsTransport(verify=verify), headers=headers)
+		login_timeout = int(options["--login-timeout"])
```
oalbrigt, line 60: This doesnt work with --disable-timeout=true (enabled by default when run from Pacemaker 2.0+). Can you add ` or 60` after the `int()` here, so it at least gets a good chance to succeed for that use-case?

thanasis2028: Ok, a check for `--disable-timeout` has been added.

agents/vmware_soap/fence_vmware_soap.py
```diff
@@ -57,7 +57,8 @@ def soap_login(options):
 
 	try:
 		headers = {"Content-Type" : "text/xml;charset=UTF-8", "SOAPAction" : "vim25"}
-		conn = Client(url + "/vimService.wsdl", location=url, transport=RequestsTransport(verify=verify), headers=headers)
+		login_timeout = 60 if "--disable-timeout" in options and options["--disable-timeout"] != "false" else int(options["--login-timeout"])
```
oalbrigt, line 60: It's easier to just do `or 60` like here:
https://github.com/ClusterLabs/fence-agents/blob/master/agents/gce/fence_gce.py#L296

As disable-timeout can be other true/false values as well.

thanasis2028: Ok, I now understand. My bad.

thanasis2028:
Note that this timeout only seems to apply to the initial connection attempt, at least with my setup. Any rpc requests can still take too long to get a reply, and cause the pacemaker to kill it (with SIGKILL). In that case, any temp files created will remain forever, as described in #446.

agents/vmware_soap/fence_vmware_soap.py
oalbrigt: No worries :).

oalbrigt:
Thanks.

PR #449, opened 2021-11-03
fence_aliyun: Optimize log output

bd233:
1. Optimize log output
2. Optimize region and ram role check logic

### DIFF
--- a/agents/aliyun/fence_aliyun.py
+++ b/agents/aliyun/fence_aliyun.py
@@ -1,52 +1,66 @@
 #!@PYTHON@ -tt
 
-import sys, re
+import sys
 import logging
 import atexit
 import json
+
 sys.path.append("@FENCEAGENTSLIBDIR@")
 from fencing import *
-from fencing import fail, fail_usage, EC_TIMED_OUT, run_delay
+from fencing import fail_usage, run_delay
+
 
 try:
 	from aliyunsdkcore import client
 	from aliyunsdkcore.auth.credentials import EcsRamRoleCredential
+	from aliyunsdkcore.profile import region_provider
+except ImportError as e:
+	logging.warn("The 'aliyunsdkcore' module has been not installed or is unavailable, try to execute the command 'pip install aliyun-python-sdk-core --upgrade' to solve. error: %s" % e)
+
+
+try:
 	from aliyunsdkecs.request.v20140526.DescribeInstancesRequest import DescribeInstancesRequest
 	from aliyunsdkecs.request.v20140526.StartInstanceRequest import StartInstanceRequest
 	from aliyunsdkecs.request.v20140526.StopInstanceRequest import StopInstanceRequest
 	from aliyunsdkecs.request.v20140526.RebootInstanceRequest import RebootInstanceRequest
-	from aliyunsdkcore.profile import region_provider
-except ImportError:
-	pass
+except ImportError as e:
+	logging.warn("The 'aliyunsdkecs' module has been not installed or is unavailable, try to execute the command 'pip install aliyun-python-sdk-ecs --upgrade' to solve. error: %s" % e)
+
 
 def _send_request(conn, request):
+	logging.debug("send request action: %s" % request.get_action_name())
 	request.set_accept_format('json')
 	try:
 		response_str = conn.do_action_with_exception(request)
-		response_detail = json.loads(response_str)
-		logging.debug("_send_request reponse: %s" % response_detail)
-		return response_detail
 	except Exception as e:
-		fail_usage("Failed: _send_request failed: %s" % e)
+		fail_usage("Failed: send request failed: Error: %s" % e)
+
+	response_detail = json.loads(response_str)
+	logging.debug("reponse: %s" % response_detail)
+	return response_detail
 
 def start_instance(conn, instance_id):
+	logging.debug("start instance %s" % instance_id)
 	request = StartInstanceRequest()
 	request.set_InstanceId(instance_id)
 	_send_request(conn, request)
 
 def stop_instance(conn, instance_id):
+	logging.debug("stop instance %s" % instance_id)
 	request = StopInstanceRequest()
 	request.set_InstanceId(instance_id)
 	request.set_ForceStop('true')
 	_send_request(conn, request)
 
 def reboot_instance(conn, instance_id):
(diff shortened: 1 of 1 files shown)


bd233:
> Can one of the admins verify this patch?

yes

agents/aliyun/fence_aliyun.py
```diff
@@ -58,20 +72,22 @@ def get_status(conn, instance_id):
...
 	request.set_PageSize(100)
 	response = _send_request(conn, request)
-	instance_status = None
 	if response is not None:
 		instance_list = response.get('Instances').get('Instance')
 		for item in instance_list:
 			instance_id = item.get('InstanceId')
-			instance_name = item.get('InstanceName')
-			result[instance_id] = (instance_name, None)
+			host_name = item.get('HostName')
```
oalbrigt, line 84: I think we'll have to do this as a try/except (EncodingError or whatever the exception it throws is) to avoid breaking functionality for current users who might not have same instance name and host name.

agents/aliyun/fence_aliyun.py
```diff
@@ -1,52 +1,66 @@
...
+try:
 	from aliyunsdkecs.request.v20140526.DescribeInstancesRequest import DescribeInstancesRequest
 	from aliyunsdkecs.request.v20140526.StartInstanceRequest import StartInstanceRequest
 	from aliyunsdkecs.request.v20140526.StopInstanceRequest import StopInstanceRequest
 	from aliyunsdkecs.request.v20140526.RebootInstanceRequest import RebootInstanceRequest
-	from aliyunsdkcore.profile import region_provider
-except ImportError:
-	pass
+except ImportError as e:
+	fail_usage("The 'aliyunsdkecs' module has been not installed or is unavailable, try to execute the command 'pip install aliyun-python-sdk-ecs --upgrade' to solve. error: %s" % e)
```
oalbrigt, line 27: It should pass here to avoid having to install all the libraries when doing builds.

agents/aliyun/fence_aliyun.py
```diff
@@ -153,18 +165,23 @@ def main():
...
+		secret_key = options["--secret-key"]
+		conn = client.AcsClient(access_key, secret_key, region)
+	elif "--ram-role" in options:
+		ram_role = options["--ram-role"]
+		role = EcsRamRoleCredential(ram_role)
+		conn = client.AcsClient(region_id=region, credential=role)
+	else:
+		fail_usage("Failed: User credentials are not set. Please set the Access Key and the Secret Key, or configure the RAM role.")
+
+	region_provider.modify_point('Ecs', region, 'ecs.%s.aliyuncs.com' % region)
```
oalbrigt, line 183: This should be in an if, so we dont fail on outputting metadata during e.g. build or if some distro provides the agent but user has to install the libraries to use it.

agents/aliyun/fence_aliyun.py
bd233: Sorry, I’m wondering,  what is the impact of users using different instance name and host name?

agents/aliyun/fence_aliyun.py
bd233: Okay, but can I use logging.error to output this message to remind users?

oalbrigt: or .warn().

agents/aliyun/fence_aliyun.py
oalbrigt: It will stop working if they got it setup with instance name and they upgrade and all of a sudden it doesnt work with instance names anymore.

bd233: Thank you for your answer, I think I understand what you mean, but I have another question: how does the instance_name  cause the user to stop working before and after the upgrade?

In addition, this API will definitely return hostname after the upgrade, and there is no failure to obtain it.

oalbrigt: Multiple nodes could have the same hostname. E.g. if you have instance name cluster1-node1 with hostname node1 and instance cluster2-node1 with node1 as hostname (probably different users using it in that case)

bd233: Yes, the same hostname is possible.
But I think in this scenario `item.get('HostName')` will not raise an exception, it will only generate a map like this:
`{"instance-id-1": ("code1", None), "instance-id-2": ("code1", None)}`, try/except is not working.

Or will such a map containing the same hostname cause errors in subsequent executions? (If yes, can you provide more details, thanks)

oalbrigt: Is it possible to use item.get('InstanceId').decode() or .encode() to solve the issue?

If not we should make a parameter called name_lookup or similar that defaults to instance and can be changed to hostname.

And we can also do the .get() in a try/except that informs users about the parameter if it fails.

bd233: Maybe I didn’t describe clearly the problem we encountered, I have restored the code here and created an issue (https://github.com/ClusterLabs/fence-agents/issues/452#issue-1053209050)
If necessary, I will solve this problem through a new PR.

agents/aliyun/fence_aliyun.py
bd233: I added a try/except, please help review whether this is appropriate

agents/aliyun/fence_aliyun.py
```diff
@@ -153,18 +165,27 @@ def main():
...
-		region = options["--region"]
-		if "--access-key" in options and "--secret-key" in options:
-			access_key = options["--access-key"]
-			secret_key = options["--secret-key"]
-			conn = client.AcsClient(access_key, secret_key, region)
-		elif "--ram-role" in options:
-			ram_role = options["--ram-role"]
-			role = EcsRamRoleCredential(ram_role)
-			conn = client.AcsClient(region_id=region, credential=role)
+	if "--region" not in options or options["--region"] == "":
```
oalbrigt, line 168: You need to change the required field for the `region` parameter, as this test would break some of the actions that doesnt require any of the required parameters (e.g. metadata).

bd233: Yes, I was ill-conceived, I restored this code

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #458, opened 2021-12-07
fence_gce: Add timeouts and failure options

kj1724:
This allows the agent to be more flexible to the users needs and improves reliability.

### DIFF
--- a/agents/gce/fence_gce.py
+++ b/agents/gce/fence_gce.py
@@ -41,6 +41,21 @@ except:
 
 METADATA_SERVER = 'http://metadata.google.internal/computeMetadata/v1/'
 METADATA_HEADERS = {'Metadata-Flavor': 'Google'}
+INSTANCE_LINK = 'https://www.googleapis.com/compute/v1/projects/{}/zones/{}/instances/{}'
+
+def run_on_fail(options):
+	if "--runonfail" in options:
+		command = options["--runonfail"]
+		logging.info("Due to failure, running command: {}".format(command))
+		os.system(command)
+
+def fail_fence_agent(options, message):
+	run_on_fail(options)
+	fail_usage(message)
+
+def raise_fence_agent(options, message):
+	run_on_fail(options)
+	raise Exception(message)
 
 #
 # Will use baremetalsolution setting or the environment variable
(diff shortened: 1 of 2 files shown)


agents/gce/fence_gce.py
```diff
@@ -41,6 +41,21 @@
 
 METADATA_SERVER = 'http://metadata.google.internal/computeMetadata/v1/'
 METADATA_HEADERS = {'Metadata-Flavor': 'Google'}
+INSTANCE_LINK = 'https://www.googleapis.com/compute/v1/projects/{}/zones/{}/instances/{}'
+
+def run_on_fail(options):
+	if "--runonfail" in options:
+		command = options["--runonfail"]
+		logging.info("Due to failure, running command: {}".format(command))
+		os.system(command)
```
oalbrigt, line 50: You should probably use `run_command()` from the fencing library instead of `os.system()` here and the other case where it's used below:
https://github.com/ClusterLabs/fence-agents/blob/5ce6098244fd2116ca1cc3a219235681c22457c8/lib/fencing.py.py#L1085

agents/gce/fence_gce.py
```diff
@@ -399,14 +438,57 @@ def define_new_opts():
...
 	}
+	all_opt["earlyexit"] = {
+		"getopt" : "",
+		"longopt" : "earlyexit",
+		"help" : "--earlyexit                    Return early if reset is already in progress",
+		"shortdesc" : "If an existing reset operation is detected, the fence agent will return before the operation completes with a 0 return code.",
+		"required" : "0",
+		"order" : 13
+	}
+	all_opt["warntimeout"] = {
```
oalbrigt, line 449: You can use "second" as type for the timeouts.

agents/gce/fence_gce.py
kj1724: Thanks! I updated the `os.system()` calls to `run_command()`

agents/gce/fence_gce.py
kj1724: Done

oalbrigt:
ok to test

oalbrigt:
Can you run `make xml-upload` and add the updated metadata to the PR?

kj1724:
I tried to run `make xml-upload` and for most of the directories it says `Nothing to be done for 'xml-upload'.` However in the agents directory it outputs a bunch of logs which looks like it should be doing something. After it is finished running none of the files change. Do I need to do anything else?

I updated the metadata file for fence_gce.

oalbrigt:
> I tried to run `make xml-upload` and for most of the directories it says `Nothing to be done for 'xml-upload'.` However in the agents directory it outputs a bunch of logs which looks like it should be doing something. After it is finished running none of the files change. Do I need to do anything else?
> 
> I updated the metadata file for fence_gce.

You probably havent installed the googleapiclient and/or oauth2client libs:
https://github.com/ClusterLabs/fence-agents/blob/64d2ef3fdac7bd4f86ebccb11282ce97d3bddbe3/configure.ac#L287-L294

oalbrigt:
Thanks.

PR #460, opened 2022-01-04
fence_lpar: Support comanaged LPARs

hramrach:
When the managed system has NovaLink installed it is still possible to
manage the system with HMC with correct chcomgmgt setting but profiles
are not supported.

@watologo1

### DIFF
--- a/agents/lpar/fence_lpar.py
+++ b/agents/lpar/fence_lpar.py
@@ -56,6 +56,18 @@ def get_power_status(conn, options):
 
 	return _normalize_status(status)
 
+def is_comanaged(conn, options):
+	conn.send("lscomgmt -m " + options["--managed"] + "\n" )
+	conn.readline()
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

oalbrigt:
retest this please

watologo1:
I cannot find any hint to what caused the build failure...?
@oalbrigt 
Any advise how to get further here would be appreciated. Thanks.

wenningerk:
> I cannot find any hint to what caused the build failure...? @oalbrigt Any advise how to get further here would be appreciated. Thanks.

Looks like coverity is stumbling over stuff in the c-code and python-code.
Although there are findings with fence_lpar.py the plethora of findings makes me
think that we are seeing some issue with the base-line - maybe triggered by renaming master to main ...

hramrach:
What are the problems reported in python code?

I seer only problems in C code in the report.

oalbrigt:
LGTM. Thanks.

PR #464, opened 2022-01-13
fence_kubevirt: take default namespace from context

dankenigsberg:
If --namespace is not provided to kubectl, a default one is taken from
kubeconfig context. Let fence_kubevirt behave similarly.

Signed-off-by: Dan Kenigsberg <danken@redhat.com>

### DIFF
--- a/agents/kubevirt/fence_kubevirt.py
+++ b/agents/kubevirt/fence_kubevirt.py
@@ -12,12 +12,21 @@ try:
 except ImportError:
     logging.error("Couldn\'t import kubernetes.client.exceptions.ApiException - not found or not accessible")
 
+def _get_namespace(options):
+    from kubernetes import config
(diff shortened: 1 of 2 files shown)


agents/kubevirt/fence_kubevirt.py
```diff
@@ -127,8 +136,6 @@ def main():
 
     run_delay(options)
 
-    validate_options(['--namespace'], options)
```
oalbrigt, line 130: Seems like you can remove the function as well (I cant see it used anywhere else).

oalbrigt:
ok to test

oalbrigt:
Can you run `make xml-upload` and attach the updated metadata to the commit?

oalbrigt:
Thanks.

PR #471, opened 2022-02-10
fence_cdu: add 8i support

watologo1:


agents/cdu/fence_cdu.py
```diff
@@ -105,20 +142,17 @@ def main():
     docs["vendorurl"] = "http://www.servertech.com"
     show_docs(options, docs)
 
-    ## Support for --plug [switch]:[plug] notation that was used before
```
oalbrigt, line 108: We shouldn't remove parameters, as that might break setups on upgrade, or people following instructions where it's used.

watologo1:
@oalbrigt Thanks for the quick review! I removed the separated patch which did remove the --switch option and did a:
git push ... --force
Should be better now?

oalbrigt (approved):

oalbrigt:
ok to test

oalbrigt:
You have to remove -- in `--api-version`, as that's only for CLI parameters, and will fail if it's not defined as an option that can be set with --api-parameter.

oalbrigt:
retest this please

oalbrigt:
Thanks.

PR #474, opened 2022-02-25
fence_lpar: fix missing import logging, use fail_usage

watologo1:
and slightly re-factor code to avoid duplicate code lines.
Should be cleanup only, no functional change.


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #475, opened 2022-03-16
fence_apc.py compatibility for Firmware major release 7 workaround

Anckermann:
Since APC/Schneider Electric released major version 7 of the firmware, this is a workaround to keep the script using the "*5s" methods. The implementation seems to still work fine, but I did not perform extensive and detailed tests. Moreover a more robust and safe control on the major version number has to be considered.


Anckermann:
My bad, I accidentaly removed the last pull request. Moved to #476

PR #497, opened 2022-07-28
fence_sbd: improve error handling

wenningerk:
basically when using 3 disks be happy with 2 answers

As run_commands - executes a couple of commands in parallel and is happy
if more than half of them do return positive - is actually generic I put it
in lib/fencing.py.py

There is a certain amount of code-duplication between run_commands and
run_command so that we could make run_command a simple wrapper.
Basic difference between run_command and run_commands, when executing a
single command, should be how stdout & stderr are collected.
With commands that do leave D state children behind subprocess.communicate
seems to be hanging till the D state child is gone. Having leftover D state children
is ugly but waiting till eternity for them to disappear is even more.

Parallel execution and kind of "quorum-decision" could be addressed in sbd
as well. But we'd still need changes in here.
So my suggestion is kind of a 2 (actually 3) step approach.
First this patch gets it to work without a change in sbd.
Then sbd introduces parallel-execution for other commands than
messaging that executes in parallel with multiple disks today already.
This sbd improvement should still work with the parallel-execution-code in
fence_sbd but gives us the possibility to remove parallel execution
from fence_sbd again.

### DIFF
--- a/agents/sbd/fence_sbd.py
+++ b/agents/sbd/fence_sbd.py
@@ -5,7 +5,7 @@ import logging
 import os
 import atexit
 sys.path.append("@FENCEAGENTSLIBDIR@")
-from fencing import fail_usage, run_command, fence_action, all_opt
+from fencing import fail_usage, run_commands, fence_action, all_opt
 from fencing import atexit_handler, check_input, process_input, show_docs
 from fencing import run_delay
 import itertools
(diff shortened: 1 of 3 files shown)


wenningerk:
hmm ... seems as if psutil isn't as available as I thought.
let me check how that goes with subprocess ...

wenningerk:
test this please

wenningerk:
Looks as if my privileges on that repo are somewhat screwed up.
Obviously I could even merge the PR but CI doesn't run on my updates.

wenningerk:
aah yes ... last update reverts requirements to what they were before - no psutil required anymore

oalbrigt:
add to whitelist

oalbrigt:
retest this please

oalbrigt:
Thanks.

Issue #515, opened 2022-11-29
group permissions - how to configure?


BlackZork:
Hi, 

I've configured fence-libvirtd 0.4 (build on arch linux from git) on host (kvm+qemu+libvirt) with libvirtd backend and multicast listener.

I would like to restrict which machine can fence which I cannot figure out how to do it. I added "groups" section this way:

```
groups {
  group {
    ip = "<host1_ip>";
    uuid = "<host1_uuid>";
    uuid = "<host2_uuid>";
  }
  group {
    ip = "<host2_ip>";
    uuid = "<host1_uuid>";
    uuid = "<host2_uuid>";
 }
}
```

If I comment out entire groups section, then host1 and host2 sees all machines and can fence any of them.

If I uncomment groups section, then both host1 and host2 can issue ```fence_xvm -o list``` and it returns only two hosts defined in group, but any attempt to control vm (off, reboot, status) gives "Permission denied" message.

I ran fence_virtd with -d 9 and In logs I see:

fence_xvm -o list from host2 (outputs list, ok).
```
Request 6 seqno 457386 domain 
Plain TCP request
ipv4_connect: Connecting to client
ipv4_connect: Success; fd = 7
Request 6 seqno 457386 src host2_ip target 
ENTER libvirt_hostlist
Sending terminator packet
Sending response to caller...
Request 6 seqno 457386 domain 
We just did this request; dropping packet
Request 6 seqno 457386 domain 
We just did this request; dropping packet
Request 6 seqno 457386 domain 
We just did this request; dropping packet
```
fence_xvm -o status -H host1 (outputs "Permission denied")
```
Request 4 seqno 324735 domain host1
Plain TCP request
ipv4_connect: Connecting to client
ipv4_connect: Success; fd = 7
Request 4 seqno 324735 src host2_ip target host1
Sending response to caller...
Request 4 seqno 324735 domain host1
We just did this request; dropping packet
Request 4 seqno 324735 domain host1
We just did this request; dropping packet
Request 4 seqno 324735 domain host1
We just did this request; dropping packet
```

fabbione:
Hi @BlackZork,

can you please confirm that you are using latest and greatest fence-virtd from fence-agents.git and not the old one from fence-virt.git?

Thanks

fabbione:
If you can please also share the full configuration, that would be ideal.

fabbione:
@BlackZork I was able to reproduce your problem and I do have a solution that will allow you to move forward.

[...]

PR #518, opened 2022-12-09
Fence virtd groups

fabbione:
Tested with both libvirt and cpg backends.
Tested with mcast, tcp, vsock and serial listeners.

NOTE: the serial listener has both "serial" and "vmchannel" mode. I was able to test only "serial" mode as vmchannel requires a libvirt managed bridge for intercepting TCP connections and I don´t have that setup.

Updated man page for group based on all different setups.

Resolves: https://github.com/ClusterLabs/fence-agents/issues/515

### DIFF
--- a/agents/virt/common/log.c
+++ b/agents/virt/common/log.c
@@ -159,7 +159,9 @@ void
 __wrap_closelog(void)
 {
 	struct log_entry *lent;
+#ifdef DEBUG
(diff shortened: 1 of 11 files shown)


oalbrigt (approved):
LGTM.

oalbrigt:
Thanks.

PR #529, opened 2023-02-18
fence_scsi: fix registration handling if ISID conflicts

smohanan20:
**Problem:**

ISID (Initiator Session ID) which belongs to I_T Nexus, changes for RHEL based on the session ID that's used for iSCSI connections. This means that the connection to the same device can be set up with different ISID on reconnects.

If an iSCSI client establishes a connection with a device using an ISID (say X) and registers/reserves with the device using a key, the key is associated with the I_T nexus (with ISID X). Now if a device reconnects for some reason (like client reboots), it establishes a new iSCSI connection (with ISID Y) and fence agent tries to register again. However fence_scsi looks up the reservation key for the device, sees that it matches with the key and skips sending a new registration command assuming that the key belongs to it (ignoring the ISID). 

**Impact:**

This can cause the iSCSI initiator to have no access to the device since it is not registered. 

**Proposed Fix:**

To fix this, the agent needs to clear the old registration and register with the device based on what the active state of the registration/reservation is. 

1. If the client holds a reservation, the agent issues a PREEMPT request with the same key. The iSCSI target will replace the old entry with the new entry.
2. If the client does not hold the reservation, the agent issues a PREEMPT request with the same key to remove the old registration. The agent sends another REGISTER command to register with the new ISID.

Both (1) and (2) are done in accordance with SCSI spec SPC3 -  (1) Section 5.6.10.4.3 "Preempting persistent reservations and registration handling" Section and (2) Section 5.6.10.4.4 "Removing registrations".

If the ISID does not change across reboots, the fix has no impact since the preempt will insert a new entry anyway.

### DIFF
--- a/agents/scsi/fence_scsi.py
+++ b/agents/scsi/fence_scsi.py
@@ -137,12 +137,41 @@ def register_dev(options, dev):
 		for slave in get_mpath_slaves(dev):
 			register_dev(options, slave)
 		return True
-	if get_reservation_key(options, dev, False) == options["--key"]:
-		return True
+
+	# Check if any registration exists for the key already. We track this in
+	# order to decide whether the existing registration needs to be cleared.
+	# This is needed since the previous registration could be for a
+	# different I_T nexus (different ISID).
+	registration_key_exists = False
+	if options["--key"] in get_registration_keys(options, dev):
+		registration_key_exists = True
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #533, opened 2023-03-17
Added --auth_plugin option to allow usage of app credentials instead of user credentials

s0urc3c0d3:
This allows users to user Barbican to generate app creds and put them in the clouds.yml:
```
clouds:
  openstack:
    auth:
      auth_url: http://CLOUD_ENDPOINT:5000
      application_credential_id: "SOMEID"
      application_credential_secret: "SOMELARGESECRET"
    region_name: "RegionOne"
    interface: "public"
    identity_api_version: 3
    auth_type: "v3applicationcredential"
```

The app creds can have smaller perm from user and they can be easly generated by users instead of full cloud accounts

To user this feature user can run:
`$ fence_openstack --cloud openstack -n INSTANCEUUID -o list --auth_plugin v3applicationcredential`

### DIFF
--- a/agents/openstack/fence_openstack.py
+++ b/agents/openstack/fence_openstack.py
@@ -1,4 +1,4 @@
-#!@PYTHON@ -tt
+#!/usr/libexec/platform-python -tt
 
 import atexit
 import logging
@@ -7,7 +7,7 @@ import os
 
 import urllib3
 
-sys.path.append("@FENCEAGENTSLIBDIR@")
+sys.path.append("/usr/share/fence")
 from fencing import *
 from fencing import fail_usage, run_delay, source_env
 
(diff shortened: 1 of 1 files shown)


agents/openstack/fence_openstack.py
```diff
@@ -1,4 +1,4 @@
-#!@PYTHON@ -tt
+#!/usr/libexec/platform-python -tt
```
oalbrigt, line 1: Dont change these, as they're replaced to the path found by `./configure` during `make`.

agents/openstack/fence_openstack.py
```diff
@@ -251,6 +258,15 @@ def define_new_opts():
         "default": 60,
         "order": 10,
     }
+    all_opt["auth_plugin"] = {
```
oalbrigt, line 261: Use - instead of _ to stay consistent with the rest of the agent.

oalbrigt (changes requested):

oalbrigt:
You'll also have to run `make xml-upload` to update the metadata and attach it to the PR, so it doesnt fail CI when running `make xml-check`.

PR #537, opened 2023-04-17
fence_scsi: Fixes for auto-detecting devices for shared VG

arslan9292:
In the earlier version, the shared devices can be auto-populated
using the 'c' flag in 'vgs' command output when shared VG is 
configured with clvm. With the release of lvmlockd, the shared VGs 
has 's' flag under 'Attr' section. 

So as to make the auto-detection feature working for lvmlockd, 
this update will allow the fence/unfence operation succeed 
even when the 'devices' attribute is left blank or undefined.

### DIFF
--- a/agents/scsi/fence_scsi.py
+++ b/agents/scsi/fence_scsi.py
@@ -314,7 +314,7 @@ def dev_read(fail=True, opt=None):
 	return devs
 
 
-def get_clvm_devices(options):
+def get_shared_devices(options):
 	devs = []
 	cmd = options["--vgs-path"] + " " +\
 	"--noheadings " +\
@@ -324,10 +324,11 @@ def get_clvm_devices(options):
 	"--config 'global { locking_type = 0 } devices { preferred_names = [ \"^/dev/dm\" ] }'"
 	out = run_cmd(options, cmd)
 	if out["rc"]:
-		fail_usage("Failed: Cannot get clvm devices")
-	for line in out["out"].split("\n"):
-		if 'c' in line.split(":")[0]:
-			devs.append(line.split(":")[1])
+		fail_usage("Failed: Cannot get shared devices")
+	for line in out["out"].splitlines():
+		vg_attr, pv_name = line.strip().split(":")
+		if vg_attr[5] in "cs":
+			devs.append(pv_name)
 	return devs
 
 
@@ -612,7 +613,7 @@ failing."
 	options["--key"] = options["--key"].lstrip('0')
 
 	if not ("--devices" in options and options["--devices"].split(",")):
-		options["devices"] = get_clvm_devices(options)
+		options["devices"] = get_shared_devices(options)
 	else:
 		options["devices"] = options["--devices"].split(",")
 


oalbrigt:
ok to test

agents/scsi/fence_scsi.py
```diff
@@ -324,9 +324,9 @@ def get_clvm_devices(options):
 	"--config 'global { locking_type = 0 } devices { preferred_names = [ \"^/dev/dm\" ] }'"
 	out = run_cmd(options, cmd)
 	if out["rc"]:
-		fail_usage("Failed: Cannot get clvm devices")
+		fail_usage("Failed: Cannot get shared devices")
 	for line in out["out"].split("\n"):
-		if 'c' in line.split(":")[0]:
+		if 'c' in line.split(":")[0] or 's' in line.split(":")[0]:
```
nrwahl2, line 329: This matches the existing pattern so there's nothing wrong with it. To add clarity and avoid redundancy, we could do:
```
        for line in out["out"].split("\n"):
                vg_attr, pv_name = line.split(":")
                if 'c' in vg_attr or 's' in vg_attr:
                        devs.append(pv_name)
```

There's also an edge case in the *existing* code that would generate a false positive: a `'c'` character in the fifth position indicates a contiguous allocation policy, not a clustered volume group. A `'c'` in the sixth position indicates a clustered volume group. From the `vgs(8)` man page:
```
       The vg_attr bits are:

       1  Permissions: (w)riteable, (r)ead-only
       2  Resi(z)eable
       3  E(x)ported
       4  (p)artial: one or more physical volumes belonging to the volume group are missing from the system
       5  Allocation policy: (c)ontiguous, c(l)ing, (n)ormal, (a)nywhere
       6  (c)lustered, (s)hared
```

I'm not 100% sure whether that VG attr bit order is consistent on all supported OSes, but `get_clvm_devices()` is sort of a guess anyway... the user can always specify a devices list if there's a problem with the dynamic check in `get_clvm_devices()`.

If we check only the sixth position (index 5 of the string), then the code could be as follows:
```
        for line in out["out"].split("\n"):
                vg_attr, pv_name = line.split(":")
                if vg_attr[5] in 'cs':
                        devs.append(pv_name)
```

nrwahl2 (commented):
Congrats on your first contribution :) It looks good.

agents/scsi/fence_scsi.py
nrwahl2: We tweaked this a little bit after Arslan tested. I think we settled on:
```
        for line in out["out"].splitlines():
                vg_attr, pv_name = line.strip().split(":")
                if vg_attr[5] in "cs":
                        devs.append(pv_name)
        return devs
```

arslan9292: Thank you Reid for recommendations and help in getting it through.

nrwahl2 (approved):
Looks good to me now. Arslan has tested this but I have not.

oalbrigt:
Thanks.

PR #538, opened 2023-04-18
fence_gce: add earlyexit parameter to power_cycle_instance for fence_gce

megelatim:
The earlyexit parameter already exists in fence_gce, this commit adds the parameter power_cycle_instance to allow it to exit before the power cycle operation is fully complete.  In most cases this will not be used.

### DIFF
--- a/agents/aws/fence_aws.py
+++ b/agents/aws/fence_aws.py
@@ -16,13 +16,13 @@ try:
 except ImportError:
 	pass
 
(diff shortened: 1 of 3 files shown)


oalbrigt:
ok to test

oalbrigt:
You have to run `make xml-upload` and attach the updated metadata to make CI pass without errors.

PR #539, opened 2023-04-18
fence_gce: add earlyexit parameter to power_cycle_instance

megelatim:
If earlyexit is enabled then power_cycle will exit before the execution of the power cycle on the gce instance is complete.


oalbrigt:
ok to test

megelatim:
Is there anything that I need to do for the test?

oalbrigt:
No. I just left it open for a couple of days so other devs had a chance to comment.

PR #542, opened 2023-04-25
fence_ibm_powervs: improved performance

andscha:
- improved performance using less power-iaas.cloud.ibm.com API calls
- add support for reboot_cycle, method to fence (onoff|cycle) (Default: onoff)

### DIFF
--- a/agents/ibm_powervs/fence_ibm_powervs.py
+++ b/agents/ibm_powervs/fence_ibm_powervs.py
@@ -12,6 +12,8 @@ from fencing import fail, run_delay, EC_LOGIN_DENIED, EC_STATUS
 state = {
 	 "ACTIVE": "on",
 	 "SHUTOFF": "off",
+	 "HARD_REBOOT": "on",
+	 "SOFT_REBOOT": "on",
 	 "ERROR": "unknown"
 }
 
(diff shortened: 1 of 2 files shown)


oalbrigt (changes requested):
Just a couple of nitpicks that should be changed, and then you need to run `make xml-upload` and attach the updated xml-file to avoid CI failing.

agents/ibm_powervs/fence_ibm_powervs.py
```diff
@@ -238,6 +270,8 @@ def main():
 	all_opt["stonith_status_sleep"]["default"] = "2"
 	all_opt["api-type"]["default"] = "private"
 	all_opt["proxy"]["default"] = ""
+	all_opt["method"]["default"] = "onoff"
```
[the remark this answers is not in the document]
andscha: ok, commented out lines: `["proxy"]["default"]` and `["method"]["default"]`

agents/ibm_powervs/fence_ibm_powervs.py
```diff
@@ -237,7 +269,9 @@ def main():
 	all_opt["power_wait"]["default"] = "1"
 	all_opt["stonith_status_sleep"]["default"] = "2"
 	all_opt["api-type"]["default"] = "private"
-	all_opt["proxy"]["default"] = ""
+	# all_opt["proxy"]["default"] = ""
+	# all_opt["method"]["default"] = "onoff"
+	all_opt["method"]["help"] = "-m, --method=[method]          Method to fence (onoff|cycle) (Default: onoff)"
```
oalbrigt, line 274: You can remove the commented out lines here, and `all_opt["method"]["help"]` as well, as that's it's default value.

andscha: ok, removed both `all_opt["method"]` lines. But activated `all_opt["proxy"]["default"]` again, since it was there before.

oalbrigt:
ok to test

oalbrigt:
Can you run `git rebase upstream/main` and then `git push -f`? Our updated CI fails to find some of the older commits.

andscha:
Did a git rebase and forced push.

oalbrigt:
Great. Thanks.

tests/data/metadata/fence_ibm_powervs.xml
```diff
@@ -75,6 +83,7 @@
 	</parameter>
 	<parameter name="debug_file" unique="0" required="0" obsoletes="debug">
 		<getopt mixed="-D, --debug-file=[debugfile]" />
+		<content type="string"  />
```
oalbrigt, line 86: You need to remove this (`make xml-upload` should solve it).

It's a special case to avoid detecting default path for files (which might change based on values from ./configure).

andscha: Removed content type="string" from debug_file parameter.

oalbrigt (approved):

oalbrigt:
Thanks.

PR #543, opened 2023-05-04
fence_scsi: Add support for space-separated devices and update in meta-data

arslan9292:
Currently the devices associated with fence_scsi should be comma-separated. With this commit, fence_scsi will also work if the 'devices' are space-separated.

Additionally, this commit includes meta-data update:
1. For fence_scsi:
 - The 'devices' parameter is optional if the cluster is configured with clvm/lvmlock.
 - The 'devices' parameter can be comma or space-separated.

2. For fence_mpath:
 - The 'devices' parameter can be comma or space-separated.


oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #546, opened 2023-06-06
spec: Migrate to SPDX license

jfriesse:
Both Fedora and openSUSE now recommends to use SPDX shortname format for License.

This work is 1:1 based on Fedora patch made by Antonio Torres <antorres@redhat.com> and he should get all credit.


oalbrigt:
Thanks.

PR #547, opened 2023-06-08
Execute source using /bin/bash not /bin/sh

shinglerb:
Attempting to execute source in /bin/sh doesn't work if /bin/sh isn't bash

### DIFF
--- a/lib/fencing.py.py
+++ b/lib/fencing.py.py
@@ -1268,7 +1268,7 @@ def fence_logout(conn, logout_string, sleep=0):
 def source_env(env_file):
     # POSIX: name shall not contain '=', value doesn't contain '\0'
     output = subprocess.check_output("source {} && env -0".format(env_file), shell=True,
-                          executable="/bin/sh")
+                          executable="/bin/bash")
     # replace env
     os.environ.clear()
     os.environ.update(line.partition('=')[::2] for line in output.decode("utf-8").split('\0') if not re.match("^\s*$", line))


lib/fencing.py.py
```diff
@@ -1268,7 +1268,7 @@ def fence_logout(conn, logout_string, sleep=0):
 def source_env(env_file):
     # POSIX: name shall not contain '=', value doesn't contain '\0'
     output = subprocess.check_output("source {} && env -0".format(env_file), shell=True,
-                          executable="/bin/sh")
+                          executable="/bin/bash")
```
fabbione, line 1271: in that case, the bash´ism should be removed. /bin/bash is not available on all systems.

oalbrigt:
It seems like replacing `source` with `.` should fix this issue.

DEmmons:
I can confirm this still causes issues on Ubuntu 24.04, for what it's worth. I ran into it with the fence_openstack agent. Two good resolutions have been suggested, and it seems that determining which one is better needs to be informed by the intention of compatibility:

- If the intention is to continue to assume the shell is bash (as the current code does), and not worry about systems that do not have bash available, this PR is just right. while it's not always the default shell, bash is available for most *nix environments that could run these agents, and even things like OSX. Ostensibly this also helps keep the behavior predictable.
- If the intention is to run in as many POSIX systems as possible with their default shell, the suggestion of using a "." instead of the bash builtin "source" is probably the right call, though this assumes the line will be interpreted the same way in each shell. While that can be understood to be a basic goal of the POSIX standard, here we are, discussing this.
- if both of these approaches leave out compatibility for an important target environment, handling beyond a one-line change may be needed, but nothing comes to mind

To sum up, unless I've misunderstood project goals, this PR looks like the right solution.

lib/fencing.py.py
oalbrigt: It works fine if you replace "source" with "." (tested with bash, dash and sh on FreeBSD):
`output = subprocess.check_output(". {} && env -0".format(env_file), shell=True, executable="/bin/sh")`

sherrardb:
i have just come here to add my "me too" to the comment from @DEmmons. i have stumbled across the same issue with fence_openstack, and i am also eagerly awaiting a solution that doesn't involve me manually patching fencing.py. :-)

per @DEmmons third bullet

> * if both of these approaches leave out compatibility for an important target environment, handling beyond a one-line change may be needed, but nothing comes to mind
> 

if there is reluctance to enact either of the first two options, then it seems that allowing `executable` to be passed as an optional argument to `source_env()`, and with a default of `/bin/sh` should satisfy all comers while also preserving the current behavior for those for whom it is not broken.

lib/fencing.py.py
DEmmons: @fabbione In the interest of seeing this resolved, I've created an alternative PR [1] that would address this concern and implement what @oalbrigt described. This does seem to work for me as well in Ubuntu.

[1] https://github.com/ClusterLabs/fence-agents/pull/650

fabbione: @DEmmons I just allowed the PR to run in CI. If it passes, i have no issues to change it source to . and move on. I totally forgot about this issue. Sorry about that.

DEmmons: As PR #650 has been merged, and addresses the same issue, this PR #547 should be closed and not merged.

oalbrigt:
Thanks. This has been solved in a separate PR.

No idea why I was having issues initially. Maybe it was on RHEL7 (Python 2.x), which was still a requirement at that time.

PR #554, opened 2023-07-19
fencing: Add Error Message & Use EC_OK

razo7:
~~- Four error codes (EC_INVALID_PRIVILEGES, EC_STATUS_HMC, EC_FETCH_VM_UUID, and EC_BAD_ARGS) have not been used, thus they can be dismissed.~~
- Add error message for EC_GENERIC_ERROR
- Use EC_OK and return result (equals to EC_OK).

### DIFF
--- a/lib/fencing.py.py
+++ b/lib/fencing.py.py
@@ -938,7 +938,7 @@ def show_docs(options, docs=None):
 		sys.exit(0)
 
 def fence_action(connection, options, set_power_fn, get_power_fn, get_outlet_list=None, reboot_cycle_fn=None, sync_set_power_fn=None):
-	result = 0
+	result = EC_OK
 
 	try:
 		if "--plug" in options:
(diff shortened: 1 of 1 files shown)


oalbrigt:
We want to keep those. They could be used by external agents, and also we dont want to change the return code numbers as pcs and other utilities map their values to know what has happened.

razo7:
I understand. What about the `Add error message for EC_GENERIC_ERROR` commit? Should it stay with EC_OK/result commit?

oalbrigt:
Yeah. That looks good to me. I dont think it was intended for that, but shouldnt hurt.

lib/fencing.py.py
```diff
@@ -573,6 +573,7 @@ def fail_usage(message="", stop=True):
 
 def fail(error_code, stop=True):
 	message = {
+		EC_GENERIC_ERROR : "Failed: Generic error"
```
oalbrigt, line 576: You forgot a comma at the end of the line.

oalbrigt:
Thanks.

PR #556, opened 2023-08-27
Update fence_vmware_soap.py

mpetricone:
Detailed error message on VM shutdown failure. The original value, "EC_WAITING_OFF" alone can be misleading.

### DIFF
--- a/agents/vmware_soap/fence_vmware_soap.py
+++ b/agents/vmware_soap/fence_vmware_soap.py
@@ -204,6 +204,7 @@ def set_power_status(conn, options):
 			if options["--action"] == "on":
 				fail(EC_WAITING_ON)
 			else:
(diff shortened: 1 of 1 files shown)


agents/vmware_soap/fence_vmware_soap.py
```diff
@@ -204,6 +204,7 @@ def set_power_status(conn, options):
 			if options["--action"] == "on":
 				fail(EC_WAITING_ON)
 			else:
+				print(ex)
```
oalbrigt, line 207: You should use the `logging.debug()`:
https://github.com/ClusterLabs/fence-agents/blob/main/doc/fa-dev-guide.md#logging-and-failing

That way it will only give you the exception info when you use -v/--verbose.

PR #559, opened 2023-09-17
Fix broken fence_xenapi.py

peter-webbird:
Fix to make --uuid work (so that this is actually usable). Added missing --plug-separator option (to prevent error message). Added ability to bypass SSL Certificate verification using --ssl-insecure

### DIFF
--- a/agents/xenapi/fence_xenapi.py
+++ b/agents/xenapi/fence_xenapi.py
@@ -127,9 +127,14 @@ def connect_and_login(options):
 	username = options["--username"]
 	password = options["--password"]
 
+	# Allow bypassing SSL verification for hosts without SSL certificates
+	skipVerification = False
+	if "--ssl-insecure" in options:
+		skipVerification = True
+	
 	try:
 		# Create the XML RPC session to the specified URL.
-		session = XenAPI.Session(url)
+		session = XenAPI.Session(url, None, None, 0, 1, skipVerification)
 		# Login using the supplied credentials.
 		session.xenapi.login_with_password(username, password)
 	except Exception as exn:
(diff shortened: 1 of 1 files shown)


agents/xenapi/fence_xenapi.py
```diff
@@ -127,9 +127,14 @@ def connect_and_login(options):
...
 
+	# Allow bypassing SSL verification for hosts without SSL certificates
+	skipVerification = False
+	if "--ssl-insecure" in options:
+		skipVerification = True
+	
 	try:
 		# Create the XML RPC session to the specified URL.
-		session = XenAPI.Session(url)
+		session = XenAPI.Session(url, None, None, 0, 1, skipVerification)
```
oalbrigt, line 137: Session() doesnt appear to have a parameter for skipVerification:
https://github.com/ClusterLabs/fence-agents/blob/main/lib/XenAPI.py.py#L121

agents/xenapi/fence_xenapi.py
```diff
@@ -191,23 +196,42 @@ def return_vm_reference(session, options):
...
+		"help" : "--plug-separator=separator              The saparator for multiple plugs.",
+		"default" : ":",
+		"required" : "0",
+		"shortdesc" : "The saparator for multiple plugs",
+		"order" : 3}
+
 def main():
 
-	device_opt = ["login", "passwd", "port", "no_login", "no_password", "session_url", "web"]
+	device_opt = ["login", "passwd", "port", "no_login", "no_password", "session_url", "web", "ssl", "uuid", "plug_separator"]
```
oalbrigt, line 219: plug_separator is included in "port" from fencing.py, so you can remove that and it's definition above.

uuid isnt neither either, as the fencing library sets it based on the --plug/-n parameter.

oalbrigt (changes requested):
A couple of changes needed, and you need to run `make xml-upload` and add the updated metadata to the PR to avoid CI failing.

agents/xenapi/fence_xenapi.py
peter-webbird: Looks like the XenAPI library bundled in this repo is significantly outdated. The latest version from https://github.com/xapi-project/xen-api/blob/master/scripts/examples/python/XenAPI/XenAPI.py should be loaded instead.

agents/xenapi/fence_xenapi.py
peter-webbird: I couldn't get it to work without these two parameters. Maybe I'm just testing it wrong? This way works for me and I have no way of validating other solutions, so someone else might need to take over this pull request (or just delete it if you prefer).

oalbrigt: You probably need to update your systems fencing library (usually located in /usr/share/fence).

PR #565, opened 2023-10-26
fence_ovm: new fence agent for Oracle VM

kaneter:
Hello.

Adds a fencing agent for virtual machines managed by Oracle VM.

fence-ovm uses Oracle VM Manager REST API.
https://docs.oracle.com/en/virtualization/oracle-vm/3.4/developer/vmapi-rest.html

Tested with Oracle VM Manager 3.4 on Oracle Linux Server 7.9.

### DIFF
--- a/agents/ovm/fence_ovm.py
+++ b/agents/ovm/fence_ovm.py
@@ -1,0 +1,193 @@
+#!@PYTHON@ -tt
+
+import sys, time
+import requests
+from requests.adapters import HTTPAdapter
+from requests.exceptions import ConnectionError, Timeout, RequestException
+from json.decoder import JSONDecodeError
+import logging
+import atexit
+import itertools
+sys.path.append("@FENCEAGENTSLIBDIR@")
+from fencing import *
+from fencing import fail, fail_usage, run_delay, EC_GENERIC_ERROR, EC_CONNECTION_LOST, EC_TIMED_OUT
+
+state = {"RUNNING" : "on", "STOPPED" : "off", "SUSPENDED" : "off"}
+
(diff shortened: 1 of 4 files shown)


oalbrigt:
Thanks.

agents/ovm/fence_ovm.py
```diff
@@ -0,0 +1,193 @@
...
+		message = result.get("message") if type(result) is dict else None
+		if message is None:
+			logging.error("Failed: Remote returned %d for '%s' request to %s", response.status_code, method, uri)
+		else:
+			logging.error("Failed: %s: %d: %s", method, response.status_code, result["message"])
+		fail(EC_GENERIC_ERROR)
+	return result
+
+def define_new_opts():
+	all_opt["ssl_client_cert_file"] = {
```
oalbrigt, line 130: You should update this to "ssl_client_certificate_file", as it's used for stdin parsing of parameters (e.g. from Pacemaker), so using the same name here will make it easier for users to create the resource based on the parameters they used in CLI testing.

kaneter: I  understand. After modification, I will create a new PR.

fabbione: You can just update this PR. no need to create a new PR each time.

kaneter: The modification is done.
Could you review this update in this PR ?
I'm concerned that this update is not reflected in this PR.

oalbrigt: Since this was already merged you'll have to create a new PR. Fabio probably didnt notice it was merged.

PR #569, opened 2024-02-11
fence_vmware_rest: monitoring action is not detecting if the API user has sufficient right to manage the fence device.

varkoly:
The call https://{api_host}/api/vcenter/vm is subject to permission checks. If the delivered list is empty the user has no rights.


oalbrigt:
Thanks.

PR #583, opened 2024-06-03
fence_docker: Connect to Docker via a UNIX socket

### LINKED ISSUE #581: Make optional the SSL verification for Docker
I understand that the fence_docker agent has set the default value of `--ssl` to true, and there is no way to set it to false via the command line. Is this true?
```
root@playground-worker:/# fence_docker -a 172.18.0.3 -n playground-worker2
2024-06-03 12:18:15,411 ERROR: Failed. If --ssl option is used, You have to also specify: --tlscert, --tlskey and --tlscacert

2024-06-03 12:18:15,411 ERROR: Please use '-h' for usage
```

and my attempt to set it to false:
```
root@playground-worker:/# fence_docker -a 172.18.0.3 -n playground-worker2 --ssl=false
ERROR:root:Parse error: option --ssl must not have an argument

ERROR:root:Please use '-h' for usage
```

If so, would you be open to set this flag to `false` by default to make optional the TLS verification? Another option is to configure the flag to accept an argument, but this might change the user interface.

### LINKED ISSUE #582: Allow connecting to Docker via UNIX socket
Currently the only option for the `fence_docker` agent is to connect to Docker only via HTTP, meaning that you have to configure Docker to accept HTTP connections.

Would you be open to extend the agent to accept a UNIX socket as a command line argument, which would configure commands send via that socket?

I have been working on this and I am willing to make a PR.

marjus45:
With this change we extend the Docker connection options to connect via a UNIX socket. To do so we add a new command line argument --unix-socket which expects the path to the Docker socket.

Also in order for this to make sense we disable the default setting of the SSL verification.

Closes #581
Closes #582

### DIFF
--- a/agents/docker/fence_docker.py
+++ b/agents/docker/fence_docker.py
@@ -48,6 +48,8 @@ def send_cmd(options, cmd, post = False):
 	output_buffer = io.BytesIO()
 	if logging.getLogger().getEffectiveLevel() < logging.WARNING:
 		conn.setopt(pycurl.VERBOSE, True)
+	if "--unix-socket" in options:
+		conn.setopt(pycurl.UNIX_SOCKET_PATH, options["--unix-socket"])
 	conn.setopt(pycurl.HTTPGET, 1)
 	conn.setopt(pycurl.URL, url.encode("ascii"))
 	if post:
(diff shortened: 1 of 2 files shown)


oalbrigt:
You can use --ssl-insecure, so you shouldnt change the default value of SSL.

You can add a note to the longdesc of --unix-socket that it is required.

oalbrigt:
And you will have to run `make xml-upload` to update the test metadata, or our CI tests will fail.

marjus45:
>You can use --ssl-insecure, so you shouldnt change the default value of SSL. You can add a note to the longdesc of --unix-socket that it is required.

Thanks for the reply @oalbrigt , as we discussed in https://github.com/ClusterLabs/fence-agents/issues/581, I can't use both `--ssl-insecure` and run over HTTP. I tried using the `--ssl-insecure` option when using a UNIX socket but it doesn't work:
```
2024-06-03 15:12:47,700 ERROR: Connection failed
2024-06-03 15:12:47,701 ERROR: Failed: Unable to obtain correct plug status or plug is not available
```

I could make running over UNIX socket a special case and configure the `send_cmd` accordinly, but I think it's better if we support both HTTP and HTTPS connections. So as you suggested I will add an extra parameter, the `--disable-ssl`.

marjus45:
> And you will have to run make xml-upload to update the test metadata, or our CI tests will fail.

Thanks @oalbrigt , I have missed this section in the docs. When I am running `./configure` from my forked repository I get the following error:
```
configure: using source epoch 2024-06-03 12:27:59 UTC
configure: error: 
  configure was unable to determine the source tree's current version. This
  generally happens when using git archive (or the github download button)
  generated tarball/zip file. In order to workaround this issue, either use
  git clone https://github.com/ClusterLabs/fence-agents.git or use an
  official release tarball. Alternatively you can add a compatible version
  in a .tarball-version file at the top of the source tree, wipe your
  autom4te.cache dir and generated configure, and rerun autogen.sh.
```

while on the direct clone it doesn't. Do you know why this happens?

marjus45:
Thanks @oalbrigt , I updated the XML file, and changed the approach to use the `--disable-ssl` flag to connect to Docker over HTTP, which will also allow running over a UNIX socket. Can you have a look again?

oalbrigt:
Thanks.

PR #591, opened 2024-07-26
fence_sbd: get devices from SBD_DEVICE env variable if devices parameter isnt set

liangxin1300:
from environment

### DIFF
--- a/agents/sbd/fence_sbd.py
+++ b/agents/sbd/fence_sbd.py
@@ -14,6 +14,7 @@ DEVICE_INIT = 1
 DEVICE_NOT_INIT = -3
 PATH_NOT_EXISTS = -1
 PATH_NOT_BLOCK = -2
+SBD_PID_FILE = "@SBDPID_PATH@"
 
 def is_block_device(filename):
     """Checks if a given path is a valid block device
@@ -356,6 +357,34 @@ Comma separated list of sbd devices",
         "order": 200
         }
 
+
+def sbd_daemon_is_running():
+    """Check if the sbd daemon is running
+    """
+    if not os.path.exists(SBD_PID_FILE):
+        logging.info("SBD PID file %s does not exist", SBD_PID_FILE)
+        return False
+
+    try:
+        with open(SBD_PID_FILE, "r") as pid_file:
+            pid = int(pid_file.read().strip())
+    except Exception as e:
+        logging.error("Failed to read PID file %s: %s", SBD_PID_FILE, e)
+        return False
+
+    try:
+        # send signal 0 to check if the process is running
+        os.kill(pid, 0)
+    except ProcessLookupError:
+        logging.info("SBD daemon is not running")
+        return False
+    except Exception as e:
+        logging.error("Failed to send signal 0 to PID %d: %s", pid, e)
+        return False
+
+    return True
+
+
 def main():
     """Main function
     """
@@ -385,7 +414,7 @@ which can be used in environments where sbd can be used (shared storage)."
     # If not specified then read SBD_DEVICE from environment
     if "--devices" not in options:
         dev_list = os.getenv("SBD_DEVICE")
-        if dev_list:
+        if dev_list and sbd_daemon_is_running():
             options["--devices"] = ",".join(dev_list.split(";"))
         else:
             fail_usage("No SBD devices specified. \
--- a/agents/sbd/fence_sbd.py
+++ b/agents/sbd/fence_sbd.py
@@ -342,7 +342,7 @@ def define_new_opts():
         "longopt" : "devices",
         "help":"--devices=[device_a,device_b] \
 Comma separated list of sbd devices",
-        "required" : "1",
+        "required" : "0",
         "shortdesc" : "SBD Device",
         "order": 1
         }
@@ -382,10 +382,14 @@ which can be used in environments where sbd can be used (shared storage)."
     docs["vendorurl"] = ""
     show_docs(options, docs)
 
-    # We need to check if --devices is given and not empty.
+    # If not specified then read SBD_DEVICE from environment
     if "--devices" not in options:
-        fail_usage("No SBD devices specified. \
-                At least one SBD device is required.")
+        dev_list = os.getenv("SBD_DEVICE")
+        if dev_list:
+            options["--devices"] = ",".join(dev_list.split(";"))
+        else:
+            fail_usage("No SBD devices specified. \
+                    At least one SBD device is required.")
 
     run_delay(options)
 
--- a/configure.ac
+++ b/configure.ac
@@ -145,6 +145,8 @@ eval FENCETMPDIR="`eval echo ${FENCETMPDIR}`"
 AC_DEFINE_UNQUOTED(FENCETMPDIR,"$FENCETMPDIR", Where Fence agents keep state files)
 AC_SUBST(FENCETMPDIR)
 
+SBDPID_PATH=${localstatedir}/run/sbd.pid
+AC_SUBST(SBDPID_PATH)
 
 if test "x$AGENTS_LIST" = x; then
 	AC_ERROR([No agents selected])
--- a/make/fencebuild.mk
+++ b/make/fencebuild.mk
@@ -9,6 +9,7 @@ define gen_agent_from_py
 		-e 's#@''SBINDIR@#${sbindir}#g' \
 		-e 's#@''LIBEXECDIR@#${libexecdir}#g' \
 		-e 's#@''FENCETMPDIR@#${FENCETMPDIR}#g' \
+		-e 's#@''SBDPID_PATH@#${SBDPID_PATH}#g' \
 		-e 's#@''IPMITOOL_PATH@#${IPMITOOL_PATH}#g' \
 		-e 's#@''OPENSTACK_PATH@#${OPENSTACK_PATH}#g' \
 		-e 's#@''AMTTOOL_PATH@#${AMTTOOL_PATH}#g' \
--- a/tests/data/metadata/fence_sbd.xml
+++ b/tests/data/metadata/fence_sbd.xml
@@ -8,7 +8,7 @@
 		<content type="string" default="reboot"  />
 		<shortdesc lang="en">Fencing action</shortdesc>
 	</parameter>
-	<parameter name="devices" unique="0" required="1">
+	<parameter name="devices" unique="0" required="0">
 		<getopt mixed="--devices=[device_a,device_b]" />
 		<content type="string"  />
 		<shortdesc lang="en">SBD Device</shortdesc>


wenningerk:
Guess we have to carefully think if this behavior is desirable.
Since it is possible to have nodes that do support running the SBD daemon (e.g. proper watchdog-device) and others that don't I'd rather have the fence-agent behave the same way regardless on which node it is running.
On the other hand if the SBD devices are available e.g. under different names on the nodes this would be an easy way to simplify the configuration in the CIB.
What was your use-case that triggered the idea?

Klaus

liangxin1300:
>What was your use-case that triggered the idea?
- keep the same with external/sbd
- If the sbd devices change (are removed or replaced), we have to remember to change fence_sbd's parameter. That might be inconsistent with /etc/sysconfig/sbd

oalbrigt:
You have to run `make xml-upload` and attach the updated test metadata to avoid CI failing.

agents/sbd/fence_sbd.py
```diff
@@ -15,6 +15,8 @@
 PATH_NOT_EXISTS = -1
 PATH_NOT_BLOCK = -2
 
+SYSCONFIG_SBD = "/etc/sysconfig/sbd"
```
gao-yan, line 18: It'd be better to pick up the build-time detection on the "initconfdir" in configure.ac for the sysconfig path so that it'd be  friendly as well for other distributions.

gao-yan (commented):
Thanks, @liangxin1300 !

> Guess we have to carefully think if this behavior is desirable. Since it is possible to have nodes that do support running the SBD daemon (e.g. proper watchdog-device) and others that don't I'd rather have the fence-agent behave the same way regardless on which node it is running.

It still makes sense to respect explicitly specified "devices" especially since there might be the cases where some of the cluster nodes are not running sbd daemon but can be a fencing executor. But OTOH IMO, it also makes sense to be possible to omit the parameter for the normal cases where all nodes are configured to be running sbd daemon. So that by falling back to sysconfig, it'll avoid unnecessary duplication or even inconsistency with what's configured in the sysconfig.

agents/sbd/fence_sbd.py
liangxin1300: Changed

configure.ac
```diff
@@ -460,8 +460,8 @@ AM_CONDITIONAL([xvmcompat], [test "x$xvmcompat" == "xyes"])
 # Try to detect the appropriate conf dir. Several systems have both /etc/default
 # and /etc/sysconfig but latter is always primary.
 AC_ARG_VAR(initconfdir, [directory for initscripts configuration])
-if test "x$initconfdir" = x && test "x$FENCE_VIRT" = "x1"; then
-	AC_CHECK_FILE(/etc/conf.d, [initconfdir='$(sysconfdir)/conf.d}'], [# Gentoo/Arch
+if test "x$initconfdir" = x; then
```
oalbrigt, line 463: This fails on BSD (see CI error). I'm not sure if there's an equivalent there.

oalbrigt: The fix makes it fail on lots of different archs on RHEL, e.g.:
https://ci.kronosnet.org/job/fence-agents/job/fence-agents-pipeline/job/PR-591/5/pipeline-console/?selected-node=538

wenningerk:
Puh ... that is a while back but was there a reason to explicitly parse sbd-config? Not 100% sure but I guess systemd is executing it as shell-script (if not we would at least have consistent behavior) so we might do here as well. And shouldn't it even be in the environment of pacemaker and thus be inherited to the fence-agents?
To be on the safe side we could as well check if sbd is enabled on the node before we use a config-file that isn't used otherwise.
Up to now not specifying devices is caught as configuration error. Not to change that behavior we might think of some explicit content here - like 'auto'. Maybe even something that lists the nodes where the config should be taken from sbd-config ... could even be combined with explicit devices for other nodes. Maybe some brainstorming in that direction makes sense ...

configure.ac
oalbrigt: Removing the newline didnt make any difference, and also makes it harder to see that it's the closing statement.

gao-yan:
> Puh ... that is a while back but was there a reason to explicitly parse sbd-config?

SBD is not an easy thing to set up. It involves setups of watchdog & block devices, on-disk metadata, sysconfig file, CIB configuration... When the other sbd fence agent was implemented, I believe it was exactly for the purpose of easing the setup and more importantly ensuring the consistency by reading already known configuration from sysconfig rather than unnecessarily repeating them in the CIB.

>Not 100% sure but I guess systemd is executing it as shell-script

Unlikely so. It seems to do the parsing by itself:
https://github.com/systemd/systemd/blob/5acca1b88551633c91ae602351f9a23af178ce6b/src/basic/env-file.c#L22

>(if not we would at least have consistent behavior) so we might do here as well.

The python module "configparser" could be a helper. Although it was originally meant for parsing ini files, but apparently it's used by us as well such as:

https://github.com/ClusterLabs/resource-agents/blob/f2932e218b912761c4374fd592dbaa6225abe6ae/tools/nfsconvert.in#L282-L284

>And shouldn't it even be in the environment of pacemaker and thus be inherited to the fence-agents?

It'd probably be asking too much of pacemaker if it was purely for the purpose of passing on the environment variables to the fence agent? It might be more convincing if pacemaker itself should care about any of the sbd settings.

>To be on the safe side we could as well check if sbd is enabled on the node before we use a config-file that isn't used otherwise.

Sounds sensible.

>Up to now not specifying devices is caught as configuration error. Not to change that behavior we might think of some explicit content here - like 'auto'.

It might arguably be a change of behavior but it wouldn't affect existing setups. I believe all of them have always had "devices" explicitly configured. 

I'd rather not unnecessarily complicate this. This agent handles fencing via disk-based sbd. It seems very strait-forward  to me that it should try finding the device list from sysconfig  if "devices"  is not explicitly specified. Of course if SBD_DEVICE is not even configured in sysconfig, an error should be dropped.

>Maybe even something that lists the nodes where the config should be taken from sbd-config ... could even be combined with explicit devices for other nodes.

Separate sbd resources configured for different nodes with specific "devices" or without it, together with location constraints should do?

oalbrigt:
retest this please

wenningerk:
> > And shouldn't it even be in the environment of pacemaker and thus be inherited to the fence-agents?
> 
> It'd probably be asking too much of pacemaker if it was purely for the purpose of passing on the environment variables to the fence agent? It might be more convincing if pacemaker itself should care about any of the sbd settings.
> 

I guess that was a misunderstanding. I meant that the environment that is currently passed to the fence-agents by pacemaker should already contain the definitions from /etc/sysconfig/sbd. Didn't check though.

gao-yan:
> I guess that was a misunderstanding. I meant that the environment that is currently passed to the fence-agents by pacemaker should already contain the definitions from /etc/sysconfig/sbd. Didn't check though.

Ah, you are right. Somehow I forgot pacemaker already retrieves sbd sysconfig. In that case, it'd be even better to just reference the environment variable SBD_DEVICE in here.

wenningerk:
Was thinking of a smart way to check for a locally running sbd daemon.
Well one more reason why it would make sense for the sbd-binary running in cmdline-mode to talk to the sbd-binary running as daemon. Unfortunately for now there is nothing and I don't see a way of using the cmdline-tool as is to check in a generic way. Actually I'd like to move the whole logic to sbd - see if there is a daemon to talk to otherwise handle in the cmdline-tool. We could do that with the whole disk-interaction so that the fence-agent doesn't have to deal with it together with all the nastyness of hanging block-devices. Unfortunately that would be a major overhaul but it might pay off in the future to have a stable interface to the cmdline-tool and have that handle everything internally.
What stays is doing it as pacemaker does - take the pid-file and check if it is running.
Other thing would be to check with systemd if sbd is enabled.

gao-yan:
> Was thinking of a smart way to check for a locally running sbd daemon. Well one more reason why it would make sense for the sbd-binary running in cmdline-mode to talk to the sbd-binary running as daemon. Unfortunately for now there is nothing and I don't see a way of using the cmdline-tool as is to check in a generic way. Actually I'd like to move the whole logic to sbd - see if there is a daemon to talk to otherwise handle in the cmdline-tool. We could do that with the whole disk-interaction so that the fence-agent doesn't have to deal with it together with all the nastyness of hanging block-devices. Unfortunately that would be a major overhaul but it might pay off in the future to have a stable interface to the cmdline-tool and have that handle everything internally.

But as long as there should be the use cases where a node is not running sbd daemon but could be an sbd fencing executor, either fence agent or sbd CLI should be able to handle hanging block devices by itself such as https://github.com/ClusterLabs/sbd/pull/148 ?

>What stays is doing it as pacemaker does - take the pid-file and check if it is running. (We already know the sbd-binary and I don't think sbd is gonna split into daemon & client binary any time soon. Only dependency we pull in is name/location of the pid-file.) Other thing would be to check with systemd if sbd is enabled. (Has to handle the freebsd-case. But anyway we should think of a way of not touching the whole agent for a freebsd build as sbd is anyway not supported.)

Checking with systemd might be more straight-forward but of course technically less generic and more of a dependency. OTOH doing it via sbd.pid would of course require logic in configure.ac dealing with `$(runstatedir)` or `${localstatedir}/run` ...

oalbrigt:
Yeah. I guess it would make sense to check if sbd is configured and running in this case.

liangxin1300:
Add a commit:

```
fence_sbd: Check if the sbd daemon is running before using SBD_DEVICE enviroment variable
    
And add @SBDPID_PATH@ for the sbd daemon pid file path
```
Please take a look

Thanks!

agents/sbd/fence_sbd.py
```diff
@@ -385,7 +414,7 @@ def main():
     # If not specified then read SBD_DEVICE from environment
     if "--devices" not in options:
         dev_list = os.getenv("SBD_DEVICE")
-        if dev_list:
+        if sbd_daemon_is_running() and dev_list:
```
wenningerk, line 417: If nothing appropriate is coming from SBD_DEVICE we don't have to check for the process.
Thus I'd turn the check around.

agents/sbd/fence_sbd.py
```diff
@@ -356,6 +357,34 @@ def define_new_opts():
...
+def sbd_daemon_is_running():
+    """Check if the sbd daemon is running
+    """
+    if not os.path.exists(SBD_PID_FILE):
+        logging.info("SBD PID file %s does not exist", SBD_PID_FILE)
+        return False
+
+    try:
+        with open(SBD_PID_FILE, "r") as pid_file:
+            pid = int(pid_file.read().strip())
```
wenningerk, line 370: Well we're opening the file in read-only and the garbage collector will finally destroy the file-object which in turn should close the file.
Not sure what is good practice in this case but as we don't know when the garbage collector will do this we might close explicitly.

liangxin1300: The `with` statement ensures the file is closed immediately after the block is executed, so we don't need to worry about when the garbage collector will run

That's the best practice by using the `with` statement

wenningerk:
Looking good from my POV.
Guess it is fine if we wait for @oalbrigt to merge!?

oalbrigt:
Thanks.

PR #592, opened 2024-08-26
Mid: fence_mpath: Allow reservation_key in multipath.conf to be the same as reservation_key in pcmk_host_map.

HideoYamauchi:
Hi All,

Currently, fence_mpath may not work if you specify the same mapping key as the multipath.conf setting.
```
(If it works)
multipath.conf - reservation_key 0x1
pcmk_host_map - "nodeA:1"
(If it does not work)
multipath.conf - reservation_key 0x1
pcmk_host_map - "nodeA:0x1"
```

This fix converts the key specified in pcmk_host_map passed by pacemaker so that both allow the same key (hexadecimal).

This fix makes unfencing/fencing possible in both cases presented above, reducing the number of mistakes users make in setting pcmk_host_map.

Best Regards,
Hideo Yamauchi.

### DIFF
--- a/agents/mpath/fence_mpath.py
+++ b/agents/mpath/fence_mpath.py
@@ -332,6 +332,7 @@ failing."
 		fail_usage("Failed: No devices found")
 
 	options["devices"] = [d for d in re.split(r"\s*,\s*|\s+", options["--devices"].strip()) if d]
+	options["--plug"] = re.sub(r'^0+', '', re.sub(r"^0x", "", options["--plug"]))
 	# Input control END
 
 	result = fence_action(None, options, set_status, get_status)


agents/mpath/fence_mpath.py
```diff
@@ -332,6 +332,7 @@ def main():
 		fail_usage("Failed: No devices found")
 
 	options["devices"] = [d for d in re.split(r"\s*,\s*|\s+", options["--devices"].strip()) if d]
+	options["--plug"] = re.sub(r'^0+', '', re.sub(r"^0x", "", options["--plug"]))
```
oalbrigt, line 335: You can shorten this to the inner function, e.g. `r"^0x0*"` for 0 or more matches of the trailing 0 or more or `r"^0x0?"` for 0 or 1 match of it.

HideoYamauchi: Hi @oalbrigt 

Since we need to take the following cases into account, we have changed the conversion to an or condition.

```
multipath.conf - reservation_key 01
pcmk_host_map - "nodeA:01"
```
How do you think?

Best Regards,
Hideo Yamauchi.

oalbrigt: Looks good to me.

oalbrigt: Oh. It should be 0x0* if not it matches "x" 0 or more times.

HideoYamauchi: Hi @oalbrigt 

> Oh. It should be 0x0* if not it matches "x" 0 or more times.

Sorry .. I made a mistake.
I pushed again.

Many thanks,
Hideo Yamauchi.

HideoYamauchi:
Hi @oalbrigt 

> Thanks.
OKay!
Thanks!

Best Regards,
Hideo Yamauchi

PR #601, opened 2024-12-08
fence_raritan_px3: fix syntax warnings

vvidic:
python3: SyntaxWarning: invalid escape sequence '\['
troff: error: invalid base glyph 'My' in composite glyph name


oalbrigt:
Thanks.

PR #621, opened 2025-03-24
fence_sbd: Add crashdump option

liangxin1300:
The parameter tells fence agent to message a "crashdump" poison pill rather than "reset/off" through sbd device.

Following the logic from https://github.com/ClusterLabs/sbd/blob/main/agent/sbd.in#L76-L79

### DIFF
--- a/agents/sbd/fence_sbd.py
+++ b/agents/sbd/fence_sbd.py
@@ -4,6 +4,7 @@ import sys, stat
 import logging
 import os
 import atexit
+import re
 sys.path.append("@FENCEAGENTSLIBDIR@")
 from fencing import fail_usage, run_commands, fence_action, all_opt
 from fencing import atexit_handler, check_input, process_input, show_docs
@@ -147,18 +148,19 @@ def send_sbd_message(conn, options, plug, message):
 
     return (return_code, out, err)
 
-def get_msg_timeout(options):
+def get_sbd_header_timeout(options):
     """Reads the configured sbd message timeout from each device.
 
     Key arguments:
     options -- options dictionary
 
     Return Value:
-    msg_timeout (integer, seconds)
+    (msg_timeout, watchdog_timeout)
     """
 
     # get the defined msg_timeout
     msg_timeout = -1 # default sbd msg timeout
+    watchdog_timeout = -1 # default sbd watchdog timeout
 
     cmd = generate_sbd_command(options, "dump")
 
@@ -174,10 +176,26 @@ def get_msg_timeout(options):
                 logging.warning(\
                         "sbd message timeouts differ in different devices")
             # we only save the highest timeout
-            if tmp_msg_timeout > msg_timeout:
-                msg_timeout = tmp_msg_timeout
+            msg_timeout = max(msg_timeout, tmp_msg_timeout)
+        elif "watchdog" in line:
+            tmp_watchdog_timeout = int(line.split(':')[1])
+            if -1 != watchdog_timeout and tmp_watchdog_timeout != watchdog_timeout:
+                logging.warning(\
+                        "sbd watchdog timeouts differ in different devices")
+            # we only save the highest timeout
+            watchdog_timeout = max(watchdog_timeout, tmp_watchdog_timeout)
 
-    return msg_timeout
+    return msg_timeout, watchdog_timeout
+
+def get_crashdump_timeout():
+    crashdump_timeout = -1
+    sbd_opts = os.getenv("SBD_OPTS", None)
+    if sbd_opts:
+        # parse '-C N'
+        match = re.search(r'-C\s+(\d+)', sbd_opts)
+        if match:
+            crashdump_timeout = int(match.group(1))
+    return crashdump_timeout
 
 def set_power_status(conn, options):
     """send status to sbd device (poison pill)
@@ -404,7 +422,6 @@ def sbd_daemon_is_running():
 
     return True
 
-
 def main():
     """Main function
     """
@@ -469,10 +486,20 @@ which can be used in environments where sbd can be used (shared storage)."
     # we check against the defined timeouts. If the pacemaker timeout is smaller
     # then that defined within sbd we should report this.
     power_timeout = int(options["--power-timeout"])
-    sbd_msg_timeout = get_msg_timeout(options)
+    sbd_msg_timeout, sbd_watchdog_timeout = get_sbd_header_timeout(options)
     if 0 < power_timeout <= sbd_msg_timeout:
         logging.warning("power timeout needs to be \
                 greater then sbd message timeout")
+
+    action = options.get("--action", None)
+    if "--crashdump" in options and action in ("reboot", "off"):
+        target_node = options.get("--plug", "UNKNOWN")
+        crashdump_timeout = get_crashdump_timeout()
+        logging.warning("crashdump option is enabled while doing action %s on node %s", action, target_node)
+        logging.warning(
+                "sbd watchdog timeout: %d, crashdump timeout(compare to setting on target-node %s): %d",
+                sbd_watchdog_timeout, target_node, crashdump_timeout
+        )
 
     result = fence_action(\
                 None, \
--- a/agents/sbd/fence_sbd.py
+++ b/agents/sbd/fence_sbd.py
@@ -192,6 +192,7 @@ def set_power_status(conn, options):
 
     target_status = options["--action"]
     plug = options["--plug"]
+    enable_crashdump = "--crashdump" in options
     return_code = 99
     out = ""
     err = ""
@@ -200,9 +201,11 @@ def set_power_status(conn, options):
     if "on" == target_status:
         (return_code, out, err) = send_sbd_message(conn, options, plug, "clear")
     elif "off" == target_status:
-        (return_code, out, err) = send_sbd_message(conn, options, plug, "off")
+        msg = "crashdump" if enable_crashdump else "off"
+        (return_code, out, err) = send_sbd_message(conn, options, plug, msg)
     elif "reboot" == target_status:
-        (return_code, out, err) = send_sbd_message(conn, options, plug, "reset")
+        msg = "crashdump" if enable_crashdump else "reset"
+        (return_code, out, err) = send_sbd_message(conn, options, plug, msg)
 
     if 0 != return_code:
         logging.error("sending message to sbd device(s) \
@@ -357,6 +360,21 @@ Comma separated list of sbd devices",
         "order": 200
         }
 
+    all_opt["crashdump"] = {
+        "getopt" : "",
+        "longopt" : "crashdump",
+        "help" : "--crashdump                    Enable crashdump, default is disabled",
+        "required" : "0",
+        "shortdesc" : "Crashdump instead of regular fence",
+        "longdesc" : "If SBD is given a fence command, this option will perform a \
+kernel crash instead of a reboot or power-off, which on a properly configured \
+system can lead to a crashdump for analysis. \
+\nWARNING:\n \
+This is unsafe for production environments. Please use with caution \
+and for debugging purposes only.",
+        "order": 201
+        }
+
 
 def sbd_daemon_is_running():
     """Check if the sbd daemon is running
@@ -390,7 +408,7 @@ def main():
     """
     # We need to define "no_password" otherwise we will be ask about it if
     # we don't provide any password.
-    device_opt = ["no_password", "devices", "port", "method", "sbd_path"]
+    device_opt = ["no_password", "devices", "port", "method", "sbd_path", "crashdump"]
 
     # close stdout if we get interrupted
     atexit.register(atexit_handler)
--- a/agents/sbd/fence_sbd.py
+++ b/agents/sbd/fence_sbd.py
@@ -228,10 +228,12 @@ def reboot_cycle(conn, options):
 
     plug = options["--plug"]
     return_code = 99
+    enable_crashdump = "--crashdump" in options
     out = ""
     err = ""
 
-    (return_code, out, err) = send_sbd_message(conn, options, plug, "reset")
+    msg = "crashdump" if enable_crashdump else "reset"
+    (return_code, out, err) = send_sbd_message(conn, options, plug, msg)
     return not bool(return_code)
 
 def get_power_status(conn, options):
--- a/tests/data/metadata/fence_sbd.xml
+++ b/tests/data/metadata/fence_sbd.xml
@@ -114,6 +114,11 @@
 		<content type="second" default="1"  />
 		<shortdesc lang="en">Sleep X seconds between status calls during a STONITH action</shortdesc>
 	</parameter>
+	<parameter name="crashdump" unique="0" required="0">
+		<getopt mixed="--crashdump" />
+		<content type="boolean"  />
+		<shortdesc lang="en">Crashdump instead of regular fence</shortdesc>
+	</parameter>
 	<parameter name="retry_on" unique="0" required="0">
 		<getopt mixed="--retry-on=[attempts]" />
 		<content type="integer" default="1"  />


wenningerk:
This is a fairly dangerous option. Just bluntly adding the switch may be a bit too dangerous. We have for instance the the query to the fence-device for the message-timeout and a check against power-timeout. Going with the default of not setting a crashdump-timeout it will even disable the watchdog (refuse or at least a big big warning in this case). And if crashdump-timeout is set we might want to see something like message-timeout + crashdump-timeout < power-timeout.
Is the intention to just have a crash-dump or to use it together on a topology with fence_kdump?

zzhou1:
Yes, sbd crashdump is very "dangerous" as expected. However, this is intended for debugging purposes[1].
No, sbd won't work with fence_kdump 

[1] https://www.suse.com/support/kb/doc/?id=000019873

oalbrigt:
You have to add the parameter in `define_new_opts()` and then add it to `device_opt`.

Make a line  in the parameter description that starts with e.g. "WARNING:" and describes how it should be used, so we can avoid most users using it incorrectly.

wenningerk:
Possibilities to enable crashdump have always been there (if not via the fence-agent at least as reboot action and maybe by editing the fence-agent) but it is a different thing if it is being actively advertised via high-level-tooling (at least pcs will do that) or if you kind of have to know how to hack something to get it configured.
But I guess if we go with a description that has a warning and on top maybe another warning to the logs when the fence-agent actually sees such a configuration we should be on the safe side (check power-timeout against crash-timeout + msg-wait. if watchdog-timeout equals crash-timeout we don't need to issue an extra warning as the daemon will keep the watchdog ticking. if crash-timeout is undefined we need an extra prominent warning to state that we are running actually without hardware-watchdog .).

oalbrigt:
> Possibilities to enable crashdump have always been there (if not via the fence-agent at least as reboot action and maybe by editing the fence-agent) but it is a different thing if it is being actively advertised via high-level-tooling (at least pcs will do that) or if you kind of have to know how to hack something to get it configured. But I guess if we go with a description that has a warning and on top maybe another warning to the logs when the fence-agent actually sees such a configuration we should be on the safe side (check power-timeout against crash-timeout + msg-wait. if watchdog-timeout equals crash-timeout we don't need to issue an extra warning as the daemon will keep the watchdog ticking. if crash-timeout is undefined we need an extra prominent warning to state that we are running actually without hardware-watchdog .).

Good idea. Maybe use `logging.error()` followed by `fail(EC_BAD_ARGS)` in main() to fail watchdog-timeout is not equal to crash-timeout.

wenningerk:
> 
> Good idea. Maybe use `logging.error()` followed by `fail(EC_BAD_ARGS)` in main() to fail watchdog-timeout is not equal to crash-timeout.

That is a good idea. It would limit the possible configurations but all other checks could simply stay in place.

liangxin1300:
Fake code flow. 
Correct me if I misunderstand something @wenningerk @oalbrigt 
```Python
    if enable_crashdump_option:
        # Check if power-timeout is greater than crash-timeout + msg-wait
        if self.power_timeout <= (self.crash_timeout + self.msg_wait):
            warning_message = "power-timeout should be greater than crash-timeout + msg-wait."
            logging.warning(warning_message)

        # Check if watchdog-timeout equals crash-timeout
        if self.watchdog_timeout != self.crash_timeout:
            error_message = "watchdog-timeout is not equal to crash-timeout."
            logging.error(error_message)
            fail(EC_BAD_ARGS)

        # Check if crash-timeout is undefined
        if self.crash_timeout is None:
            warning_message = "running without hardware-watchdog since crash-timeout is undefined."
            logging.warning(warning_message)

```

gao-yan:
Technically a crashdump timeout is specified on the fencing target and it's meant to:
* keep the watchdog armed
* be applied as the new watchdog timeout, which may be longer than the original watchdog timeout, to allow a crashdump to complete before watchdog functions.

On the fencing executor, the fact is, an sbd "message" command anyway takes "msgwait" timeout to return. The point is, "msgwait" has to be configured also longer than any crashdump timeout specified on the fencing target, not just longer than the original watchdog timeout. Administrators have to ensure that.

FWIW,  it's not necessarily easy for the fence executor to retrieve any crashdump timeout that might be specified on the fencing target unless it assumes the same is specified on the fencing executor. And still, crashdump timeout is not even a named option. It can only appear as a "-C" daemon option...

Anyway it's sensible to issue a WARNING that is as informative as possible.

wenningerk:
Yep - you have a good point here - most of the checks we've been talking above just make sense under the bold assumption that sbd-configuration (timeouts) is the same on all nodes.
Unfortunately a fencing-resource wouldn't run as clone on all nodes usually as that we could do the checks in a monitor on all possible targets.

agents/sbd/fence_sbd.py
```diff
@@ -192,6 +192,7 @@ def set_power_status(conn, options):
 
     target_status = options["--action"]
     plug = options["--plug"]
+    enable_crashdump = options["--crashdump"] == "1"
```
oalbrigt, line 195: This should be `enable_crashdump = "--crashdump" in options`, which supports all variants of of crashdump=1, true, yes, on and --crashdump from CLI.

agents/sbd/fence_sbd.py
```diff
@@ -357,6 +360,22 @@ def define_new_opts():
         "order": 200
         }
 
+    all_opt["crashdump"] = {
+        "getopt" : "",
+        "longopt" : "crashdump",
+        "help" : "--crashdump=[0|1]              Enable crashdump (Default: 0, disabled)",
```
oalbrigt, line 366: You can remove `=[0|1]` here, so it only requires `--crashdump` to enable it from CLI, and from Pacemaker or stdin it will accept `crashdump=[1|yes|on|true]` to enable it.

Also run it with -h check if  the description lines up with the other parameters.

oalbrigt:
retest this please

gao-yan:
> Unfortunately only idea I have atm to run real checks would be cloning the fencing-resource as that we have monitors running on all potential targets. Not sure if enforcing this by locking the write of a crashdump-message for target nodes that don't have a clone running would make any sense. Any other ideas?

One of the facts is, even the more basic principle " msgwait >= 2 * watchdog-timeout" is not checked by fence agent ...

It might be arguable whether fence agent should take the responsibility especially since that'd require changes to the conventional setups. But anyways, UIs like crmsh/pcs should ensure/validate the relevant settings.

wenningerk:
> > Unfortunately only idea I have atm to run real checks would be cloning the fencing-resource as that we have monitors running on all potential targets. Not sure if enforcing this by locking the write of a crashdump-message for target nodes that don't have a clone running would make any sense. Any other ideas?
> 
> One of the facts is, even the more basic principle " msgwait >= 2 * watchdog-timeout" is not checked by fence agent ...
> 
> It might be arguable whether fence agent should take the responsibility especially since that'd require changes to the conventional setups. But anyways, UIs like crmsh/pcs should ensure/validate the relevant settings.

True - to a certain extent.
This check would be probably more a task for the sbd-daemon.
As long as one stays with defaults it will be Ok anyway.
```
timeout_msgwait            = 2 * SBD_WATCHDOG_TIMEOUT_DEFAULT;
```
Unfortunately the daemon doesn't know about the Power-Timeout.
Of course high-level-tooling could take the job but for critical stuff we're usually taking into account some complex history or direct xml-editing may have messed things up.
So in an ideal world high-level-tooling directly keeps you from adding problematic configuration and for certain critical stuff those conditions are being caught at runtime as well.

agents/sbd/fence_sbd.py
```diff
@@ -357,6 +360,21 @@ def define_new_opts():
...
+        "shortdesc" : "Crashdump instead of regular fence",
+        "longdesc" : "If SBD is given a fence command, this option will instead perform a \
+kernel crash of a reboot or power-off, which on a properly configured \
+system can lead to a crashdump for analysis. \
+\nWARNING:\n \
+This is less safe for production environments. Please use with caution \
+and for debugging purposes only.",
+        "order": 201
+        }
+
```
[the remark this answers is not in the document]
gao-yan: The topic is definitely worth a whole piece of knowledgebase article :-)

Indeed crashdump for a proactive self-reset needs to be covered by sbd sysconfig SBD_TIMEOUT_ACTION. But it doesn't really make a difference for the fencing executor regarding msgwait.

Thinking of a bad enough situation in the case of poison pill, where the fencing target barely receives the poison pill right before the original watchdog-timeout pops up. Then it starts crashdump and it supposedly  takes crashdump-timeout for watchdog to trigger reset.

The basic principle is for msgwait to be already twice the watchdog-timeout so that it has enough margin for overheads. To be safe, a crashdump timeout might need to be added to it, so probably:

 msgwait >= 2 * watchdog-timeout + crashdump-timeout

?

Limited by the lengh, a warning could be something like:

"
Beware, by defaulting "-C" option for sbd daemon on the fencing target, it will disarm watchdog and do crashdump without a guaranteed reboot. It's strongly recommended to always keep watchdog armed by specifying a  crashdump timeout via "-C" option for sbd daemon on every fencing target and make sure `msgwait >= 2 * watchdog-timeout + crashdump-timeout`. Corresponding adjustments to the power timeout and stonith-timeout settings may be needed."
"
?

agents/sbd/fence_sbd.py
```diff
@@ -357,6 +360,21 @@ def define_new_opts():
...
         }
 
+    all_opt["crashdump"] = {
+        "getopt" : "",
+        "longopt" : "crashdump",
+        "help" : "--crashdump                    Enable crashdump, default is disabled",
+        "required" : "0",
+        "shortdesc" : "Crashdump instead of regular fence",
+        "longdesc" : "If SBD is given a fence command, this option will instead perform a \
+kernel crash of a reboot or power-off, which on a properly configured \
```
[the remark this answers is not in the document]
liangxin1300: Changed as `this option will perform a kernel crash instead of a reboot or power-off`

wenningerk:
> > Unfortunately the daemon doesn't know about the Power-Timeout.
> 
> IMO, given that fence_sbd knows the exact value of `msgwait`, it should probably automatically apply an appropriate power timeout without requiring users to specify/fix the setting so that it allows "message" command to take as long as it at least needs. We may consider it a calculated default or a practical value.

Valid argument and it triggers some kind of dejavue-feeling - any memory why we didn't do that @oalbrigt?
Always a question where to add automation into the chain watchdog-timeout / stonith-watchdog-timeout / msg-timeout / power-timeout / stonith-timeout ... and where we just give recommendations and otherwise warn or lock - depending on risk.

wenningerk:
Looking good!

Unfortunately there isn't much we can do about real checking without significant effort.

If we log a similar warning as in the description with each fencing-attempt using crashdump that wouldn't unnecessarily fill up the logs but it still would be valuable for later analysis to see that the cluster was running in a debugging / non production  configuration at that time. 
```
if "--crashdump" in options and options.get("--action", None) == ["reboot", "off"]:
...
```
just before the ```result =``` line

We could add the current value of watchdog-timeout and crashdump-timeout on the node that triggered fencing. like ```fencing node has watchdog-timeout=... crashdump-timeout=... - check for consistency on the target node``` or similar.

liangxin1300:
> We could add the current value of watchdog-timeout and crashdump-timeout on the node that triggered fencing. like `fencing node has watchdog-timeout=... crashdump-timeout=... - check for consistency on the target node` or similar.

I want to confirm:
- watchdog-timeout comes from `sbd -d <SBD_DEVICE> dump`
- crashdump-timeout comes from value of sbd -C option (if configured in /etc/sysconfig/sbd)?
```
-C N
           Watchdog  timeout  to  set  before crashdumping.
```
Am I understand clearly?

wenningerk:
We could modify ```get_msg_timeout``` into something that gets an n-tuple of parameters from the disk-header. Then we don't have an additional sbd-command to execute. And that should already be a cluster-wide config shared via the disk-header so no reference to check on target needed there.
Basically the crashdump-timeout comes out of /etc/sysconfig/sbd (or wherever the distro stores that information) but actually I'd expect the environment variables already be set. Would it be interesting to dump the reboot-action as well here? If the message goes through it would be irrelevant but if not that would be what we expect. Both of cause just local info and reference to check on the target may be adequate.

liangxin1300:
Add commit: 
```
    fence_sbd: Dump more information when crashdump option is enabled
    When the crashdump option is enabled, the fence_sbd agent will dump:
    - the current action
    - sbd watchdog timeout
    - sbd crashdump timeout
```

agents/sbd/fence_sbd.py
```diff
@@ -15,6 +17,7 @@
 PATH_NOT_EXISTS = -1
 PATH_NOT_BLOCK = -2
 SBD_PID_FILE = "@SBDPID_PATH@"
+SBD_SYSCONFIG_FILE = "/etc/sysconfig/sbd"
```
oalbrigt, line 20: Use @INITCONFDIR@ instead of /etc/sysconfig here:
https://github.com/ClusterLabs/fence-agents/pull/622

agents/sbd/fence_sbd.py
```diff
@@ -404,6 +430,32 @@ def sbd_daemon_is_running():
...
 
+def unquote(s):
+    """
+    Reverse shell-quoting a string, so the string '"a b c"'
+    becomes 'a b c'
+    """
+    sp = shlex.split(s)
+    return sp[0] if sp else ""
+
+def parse_sysconfig(sysconfig_file):
```
wenningerk, line 441: systemd ```EnvironmentFile=``` doesn't have the file executed by a shell but does parsing rather.
So a lot of stuff that would require a shell like variable-references aren't working anyway.
But things like special meaning of a ```\``` are implemented in systemd. Anyway did you check
if we need to do that here at all or if we can use the environment inherited from pacemaker so
that we wouldn't have to deal with checking for compatibility in parsing with systemd. Sry didn't
have the chance to check myself since my last comments.

agents/sbd/fence_sbd.py
```diff
@@ -15,6 +17,7 @@
 PATH_NOT_EXISTS = -1
 PATH_NOT_BLOCK = -2
 SBD_PID_FILE = "@SBDPID_PATH@"
+SBD_SYSCONFIG_FILE = "@INITCONFDIR@/sbd"
```
wenningerk, line 20: Do we need it? See comment with config-file parser.

agents/sbd/fence_sbd.py
```diff
@@ -469,11 +521,17 @@ def main():
...
+    sbd_msg_timeout, sbd_watchdog_timeout = get_sbd_header_timeout(options)
     if 0 < power_timeout <= sbd_msg_timeout:
         logging.warning("power timeout needs to be \
                 greater then sbd message timeout")
 
+    action = options.get("--action", None)
+    if "--crashdump" in options and action in ("reboot", "off"):
+        crashdump_timeout = get_crashdump_timeout()
+        logging.warning("crashdump option is enabled while doing %s action", action)
+        logging.warning("sbd watchdog timeout: %d, crashdump timeout: %d", sbd_watchdog_timeout, crashdump_timeout)
```
wenningerk, line 533: maybe 'crashdump timeout (compare to setting on target-node %s): ...' to make it clear that we actually don't know about the relevant setting.

agents/sbd/fence_sbd.py
liangxin1300: Removed `parse_sysconfig`, reading environment value `SBD_OPTS` instead

agents/sbd/fence_sbd.py
liangxin1300: Removed

agents/sbd/fence_sbd.py
liangxin1300: Add target node(from reading --plug option) here

wenningerk: yeah - target node is interesting in any case although it might be logged in the course of the fencing action already. My suggestion was as well intended to make it obvious which config parameters are synced (watchdog-timeout) or potentially configured individually (crash-timeout, reboot-action, ...) on each node (and thus we advise to compare with settings on the target node).

liangxin1300: I think the current log already shows enough details.
Recently, I was also working on the interface to set crashdump from crmsh by using such interface `crm sbd configure crashdump-watchdog-timeout=<int>`, then crmsh will make sure all settings are synced between cluster nodes.

liangxin1300: > maybe 'crashdump timeout (compare to setting on target-node %s): ...' to make it clear that we actually don't know about the relevant setting.

Changed

agents/sbd/fence_sbd.py
```diff
@@ -449,11 +486,21 @@ def main():
...
                 greater then sbd message timeout")
 
+    action = options.get("--action", None)
+    if "--crashdump" in options and action in ("reboot", "off"):
+        target_node = options.get("--plug", "UNKNOWN")
+        crashdump_timeout = get_crashdump_timeout()
+        logging.warning("crashdump option is enabled while doing action %s on node %s", action, target_node)
+        logging.warning(
+                "sbd watchdog timeout: %d, crashdump timeout(compare to setting on target-node %s): %d",
+                sbd_watchdog_timeout, target_node, crashdump_timeout)
```
oalbrigt, line 501: You got an unmatched `)` here.

liangxin1300:
Can this PR be merged?

Thanks!

gao-yan:
> > > Unfortunately the daemon doesn't know about the Power-Timeout.
> > 
> > 
> > IMO, given that fence_sbd knows the exact value of `msgwait`, it should probably automatically apply an appropriate power timeout without requiring users to specify/fix the setting so that it allows "message" command to take as long as it at least needs. We may consider it a calculated default or a practical value.
> 
> Valid argument and it triggers some kind of dejavue-feeling - any memory why we didn't do that @oalbrigt? Always a question where to add automation into the chain watchdog-timeout / stonith-watchdog-timeout / msg-timeout / power-timeout / stonith-timeout ... and where we just give recommendations and otherwise warn or lock - depending on risk.

Generally, I think as long as the previous level of timeout is clearly known to the current context,  we should better allow it to time out rather than interrupt unless there's a good reason.

For instance, I made pacemaker tolerate any misconfiguration of sthonith-timeout that is shorter than stonith-watchdog-timeout, basically because pacemaker clearly knows the value of stonith-watchdog-timeout:
https://github.com/ClusterLabs/pacemaker/pull/3385

But in the case of poison pill, pacemaker cannot do the similar for msgwait timeout since pacemaker doesn't know it.

In this case for fence_sbd, I think we'd better at least apply a calculated default for power-timeout if the setting is omitted. And even, if there's any explicit power-timeout setting that is too short, we'd probably better tolerate it by applying a reasonable value instead. That'd be a big favor to users :-)

liangxin1300:
> In this case for fence_sbd, I think we'd better at least apply a calculated default for power-timeout if the setting is omitted. And even, if there's any explicit power-timeout setting that is too short, we'd probably better tolerate it by applying a reasonable value instead. That'd be a big favor to users :-)

fence_sbd already set power-timeout a default value as 30s [1];
And also raise a warning if it's shorter than msg timeout [2].
[1] https://github.com/ClusterLabs/fence-agents/blob/main/agents/sbd/fence_sbd.py#L402
[2] https://github.com/ClusterLabs/fence-agents/blob/main/agents/sbd/fence_sbd.py#L453-L455

gao-yan:
> > In this case for fence_sbd, I think we'd better at least apply a calculated default for power-timeout if the setting is omitted. And even, if there's any explicit power-timeout setting that is too short, we'd probably better tolerate it by applying a reasonable value instead. That'd be a big favor to users :-)
> 
> fence_sbd already set power-timeout a default value as 30s [1]; And also raise a warning if it's shorter than msg timeout [2]. [1] https://github.com/ClusterLabs/fence-agents/blob/main/agents/sbd/fence_sbd.py#L402 [2] https://github.com/ClusterLabs/fence-agents/blob/main/agents/sbd/fence_sbd.py#L453-L455

Yes, I knew :-)

I mean if the default power-timeout (30s) or a configured value by user is shorter than our calculated value which is based on msgwait, use our calculated value.

oalbrigt:
We have some CI issues today it seems like, so just ignore that for now.

oalbrigt:
retest this please

wenningerk:
Do we actually adapt stonith-timeout if power-timeout is too long?
If not this automatism would lead to things falling apart at a non-expected point if stuff read from the disk-header changes. Usually it is preferable if failures are displayed/happen as close as possible to where a change happened
and as reproducible as possible.
We have this automatism to adhere the longest msg-timeout we read from disk. In a 3 disk-setup we would still
fence if we see just 2 disks. Is there any risk with inconsistent disk-headers? I mean the fencing side seeing 2 disks while the target-side is seeing a different set of 2 disks with the one with the large msg-wait being seen just by  the target-side. Not that prior to this automatism we would be safe in all cases but it might obfuscate the mis-configuration.
Anyway I'm trying to be a bit careful with these automatisms that might obfuscate a mis-configuration that they still might not cure entirely but just move the point where things are going wrong. So basically I prefer failing instead of trying to be more intelligent than the admin. A warning in a fencing-test might be overseen a failure will not.
Not saying automatism in this case doesn't make sense or is dangerous I'd just like to think it through carefully.
Lessons learned from stonith-watchdog-timeout=-1 ...

gao-yan:
> Do we actually adapt stonith-timeout if power-timeout is too long?

Well, it turns out now that it doesn't even need to bother taking care of power-timeout if invoked by pacemaker, at least by default: https://github.com/ClusterLabs/fence-agents/pull/357

I really favor the current behavior that by default power-timeout is basically disabled so that users don't need to take care of this level of timeout.

OTOH, do you mean we'd want fence_sbd to accordingly adjust stonith-timeout like the LHA sbd agent does:

https://github.com/ClusterLabs/sbd/blob/59461195898396799e5b435c0394e240d9a0aca9/agent/sbd.in#L61
?

> If not this automatism would lead to things falling apart at a non-expected point if stuff read from the disk-header changes.
> Usually it is preferable if failures are displayed/happen as close as possible to where a change happened and as reproducible as possible.
> We have this automatism to adhere the longest msg-timeout we read from disk. In a 3 disk-setup we would still fence if we see just 2 disks. Is there any risk with inconsistent disk-headers? I mean the fencing side seeing 2 disks while the target-side is seeing a different set of 2 disks with the one with the large msg-wait being seen just by the target-side. Not that prior to this automatism we would be safe in all cases but it might obfuscate the mis-configuration.
> Anyway I'm trying to be a bit careful with these automatisms that might obfuscate a mis-configuration that they still might not cure entirely but just move the point where things are going wrong. So basically I prefer failing instead of trying to be more intelligent than the admin. A warning in a fencing-test might be overseen a failure will not.
> Not saying automatism in this case doesn't make sense or is dangerous I'd just like to think it through carefully. Lessons learned from stonith-watchdog-timeout=-1 ...

I understand your concerns about such a misconfiguration.

It'd make sense for the agent to try finding any inconsistent watchdog/msgwait timeouts between the disks and anyway warn users about that, or even refuse to fence if to be completely safe...  But that should be a different improvement.

The thing is, it's anyway not an ideal way to rely on power-timeout to spot such a misconfiguration...

But since power-timeout is basically disabled by default when an agent is invoked by pacemaker, I don't really mind whether we automatically calculate it or now :-)

wenningerk:
> > Do we actually adapt stonith-timeout if power-timeout is too long?
> 
> Well, it turns out now that it doesn't even need to bother taking care of power-timeout if invoked by pacemaker, at least by default: #357
> 
> I really favor the current behavior that by default power-timeout is basically disabled so that users don't need to take care of this level of timeout.

Yep tackling the power-timeout issue had given me some dejavu feeling.
Thanks for bringing that to my mind ;-)

> 
> But since power-timeout is basically disabled by default when an agent is invoked by pacemaker, I don't really mind whether we automatically calculate it or not :-)

Anyway I was a bit surprised when we started taking this into the crashdump PR.
So my suggestion would be to move it to a separate PR (or store it in some safe place ;-) )
and get this one merged.

liangxin1300:
> Add a new commit:
> 
> ```
>     fence_sbd: Adjust power timeout
>     
>     Adjust power timeout when it is less or equal to SBD message timeout.
>     Following the formula: power_timeout = int(1.2 * sbd_timeout)
> ```

Removed it

oalbrigt:
Thanks.

PR #624, opened 2025-04-17
fence_virtd: fix linkage of server modules

fundawang:
linking libs should be put into `LIBADD` rather than `LDFLAGS` for shared libraries, including modules. Otherwise it may cause [underlink problems](https://wiki.mageia.org/en/Underlinking_issues_in_packaging#How_to_fix) with older libtool.

Another problem is `virt` and `cpg` are actually using libuuid, but the library is not linked.


oalbrigt:
Thanks.

PR #650, opened 2026-04-15
fencing.py: change from Bash "source" to POSIX-compliant "."

DEmmons:
In the source_env function, the Bash-specific "source command is used, but the executable is defined as "/bin/sh", which in many systems may not evaluate to Bash. A PR exists to address this [1] by ensuring "/bin/bash" is the executable used, but the reviewer request to implement a more POSIX-compliant approach and not require Bash on the system seems to be unanswered for a couple years. This tiny PR would do just that.

[1] https://github.com/ClusterLabs/fence-agents/pull/547

### DIFF
--- a/lib/fencing.py.py
+++ b/lib/fencing.py.py
@@ -1274,7 +1274,7 @@ def fence_logout(conn, logout_string, sleep=0):
 
 def source_env(env_file):
     # POSIX: name shall not contain '=', value doesn't contain '\0'
-    output = subprocess.check_output("source {} && env -0".format(env_file), shell=True,
+    output = subprocess.check_output(". {} && env -0".format(env_file), shell=True,
                           executable="/bin/sh")
     # replace env
     os.environ.clear()


oalbrigt:
From what I remember this didnt work in my testing back then. I'll test again next week on RHEL8 and 9 to see if it works fine now.

DEmmons:
> From what I remember this didnt work in my testing back then. I'll test again next week on RHEL8 and 9 to see if it works fine now.

I appreciate this - I think I interpreted your comment before as "tested with bash, dash [on Linux, probably Red Hat] and sh on FreeBSD", but I take it that was all on FreeBSD. For my part I have tested it on Ubuntu, and while the existing code fails (/bin/sh is Dash in Debian and Ubuntu), changing either 'source' to '.', or 'executable="/bin/sh"' to 'executable="/bin/bash"' work, and changing both also works - indicating it should work in systems where /bin/sh points to bash as well. Hopefully testing on RHEL confirms this.

DEmmons:
For what it's worth, this does work on my test container running CentOS Stream 9

oalbrigt:
Sounds good to me. I just need to test all the scenarioes I can think of to make sure we dont break it for anyone.

oalbrigt:
Thanks.

That works well on RHEL8 and RHEL9.

Not sure if it was RHEL7 or RHEL6 I saw issues in, but we dont need to support those upstream anymore.

PR #651, opened 2026-05-05
fence_openstack: add auth-plugin support for clouds.yaml

ZhichaoTJ:
## Summary

`fence_openstack` currently forces the Keystone `password` auth plugin, including when credentials are read from `clouds.yaml`.

This change adds an `--auth-plugin` option, defaulting to `password`, and allows `clouds.yaml` `auth_type` / `auth` values to be used for non-password keystoneauth plugins such as `v3applicationcredential`.

## Motivation

In federated OpenStack environments, users may not have local Keystone passwords. Application credentials are a common non-interactive authentication mechanism for automation and HA/fencing tools.

## Compatibility

The default remains `password`, so existing username/password behavior is preserved.

This PR intentionally keeps the scope small:
- no openrc behavior changes
- no SSL/cacert behavior changes
- no unrelated refactoring

## Testing

- `python -m py_compile agents/openstack/fence_openstack.py`: passed
- `./autogen.sh` / `./configure` / `make`: not completed locally on macOS; `configure` failed because Apple linker does not support `-Wl,--enable-new-dtags`

### DIFF
--- a/agents/openstack/fence_openstack.py
+++ b/agents/openstack/fence_openstack.py
@@ -299,9 +299,7 @@ def main():
     all_opt["port"]["shortdesc"] = "UUID of the node to be fenced."
     all_opt["power_timeout"]["default"] = "60"
 
-    input_options = process_input(device_opt)
-    auth_plugin_provided = "--auth-plugin" in input_options
-    options = check_input(device_opt, input_options)
+    options = check_input(device_opt, process_input(device_opt))
 
     # workaround to avoid regressions
     if "--uuid" in options:
@@ -329,7 +327,7 @@ This agent calls the python-novaclient and it is mandatory to be installed "
     auth_options = None
     if options.get("--cloud"):
         cloud = get_cloud(options)
-        if not auth_plugin_provided:
+        if options["--auth-plugin"] == "password":
             options["--auth-plugin"] = cloud.get("auth_type") or options["--auth-plugin"]
         if options["--auth-plugin"] != "password":
             auth_options = cloud.get("auth")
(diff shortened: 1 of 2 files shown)


ZhichaoTJ:
I noticed an earlier PR attempted similar support. I tried to keep this version minimal and address the previous review feedback by keeping the generated shebang unchanged and using `--auth-plugin` naming.

agents/openstack/fence_openstack.py
```diff
@@ -284,7 +299,9 @@ def main():
     all_opt["port"]["shortdesc"] = "UUID of the node to be fenced."
     all_opt["power_timeout"]["default"] = "60"
 
-    options = check_input(device_opt, process_input(device_opt))
+    input_options = process_input(device_opt)
```
oalbrigt: There's no reason to change this part of the code.

You can use e.g. `if "--auth-plugin" in options" to check for this without changing the processing.

ZhichaoTJ: Thanks for the review. I updated the PR accordingly,

ZhichaoTJ:
Hi, the PR builds on RHEL 9 and has been validated in our environment.

It looks like Jenkins is currently blocked on project admin authorization rather than an actual build failure. Could one of the project admins please authorize or retrigger the CI run?

oalbrigt:
You need to run `make xml-upload` and add the updated metadata to the PR for CI to pass `make xml-check`.

ZhichaoTJ:
> You need to run `make xml-upload` and add the updated metadata to the PR for CI to pass `make xml-check`.

Thanks, that was exactly the missing step. I regenerated the metadata with make xml-upload and pushed a follow-up commit updating tests/data/metadata/fence_openstack.xml. The PR should now be aligned for xml-check.

oalbrigt:
Thanks.

PR #655, opened 2026-06-16
fence_aws: add tag/eni/ebs identity-method instance resolution

zypek:
## What

Adds an `--identity-method` option to `fence_aws` so a node can be resolved to its EC2 instance at fence time by something other than a static instance ID:

- `instance-id` (default) — unchanged behaviour, `--plug` is the instance ID
- `tag` — `--plug` is the value of a unique EC2 tag (`--tag=<key>`, e.g. `Name`)
- `eni` — `--plug` is an ENI id; the agent resolves the attached instance
- `ebs` — `--plug` is an EBS volume id; the agent resolves the attached instance

## Why

When an instance is replaced — Auto Scaling, blue/green — its instance ID changes, so a STONITH device keyed on the old ID can no longer fence that node. Resolving the target by a stable tag, or by a persistent ENI/EBS volume that follows the workload, keeps fencing working across replacement without rewriting the CIB.

## Behaviour

- The fencing library learns target state only through `get_power_status()` (the pre-check and the post-"off" poll). With tag/eni/ebs identity a terminated instance no longer resolves, so `get_power_status()` distinguishes **"target confirmed dead → off"** from **"cannot determine → fail"**. This is what allows an already-terminated node (e.g. replaced by Auto Scaling) to be fenced successfully. `set_power_status()` fails safe when it cannot resolve a target — it never reports a false-positive fence. The `instance-id` path is unchanged.
- `region` is now a required option; it was silently defaulting before. A missing region now fails clearly, and `pcs`/Pacemaker rejects an incomplete stonith config.
- IMDS access uses IMDSv2 with explicit connect/read timeouts so the fence path never blocks on an unreachable metadata service; the agent's own instance-id/build_number are read once at startup instead of on every call.
- Tag/ENI/EBS lookups retry with backoff for AWS eventual consistency. Multiple instances sharing a tag are disambiguated by a `build_number` tag, refusing to guess rather than risk fencing the wrong node.
- `stopping`/`shutting-down`/`terminated` now map to `off` so a fence of an already-stopping or terminated instance completes.

## Testing

- `make`, `make xml-check`, and `make check` (rng-check / delay-check) pass; metadata validates against `lib/metadata.rng`. `tests/data/metadata/fence_aws.xml` is regenerated for the new options.
- Exercised on a 5-node RHEL 8.10 / Pacemaker 2.1.7 cluster: 40/40 successful fences across all four identity methods (instance-id, tag, ebs, eni), including failover of a promotable resource each round.

Earlier revisions of these changes were reviewed by Oyvind Albrigtsen.

### DIFF
--- a/agents/aws/fence_aws.py
+++ b/agents/aws/fence_aws.py
@@ -3,6 +3,7 @@
 import sys, re
 import logging
 import atexit
+import time
 sys.path.append("@FENCEAGENTSLIBDIR@")
 from fencing import *
 from fencing import fail, fail_usage, run_delay, EC_STATUS, SyslogLibHandler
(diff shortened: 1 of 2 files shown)


oalbrigt:
Thanks.

PR #656, opened 2026-06-17
fence_openstack: fix list-action to avoid timeout when there are 100+ VMs on the hypervisor

arslan9292:
Routine `monitor` actions internally translate to a full `list` command.
On large deployments, fetching all VMs from the Nova API creates massive
payloads that cause the agent to timeout.

Changes in this patch:
* Fetch only 1 VM (`limit=1`) during `monitor` operations to eliminate
API overhead and prevent timeouts.


oalbrigt:
Thanks.

PR #657, opened 2026-07-01
fence_openstack: implement reboot_cycle_fn for hard reboot support

adyan1025:
fence_openstack has server.reboot("HARD") code in set_power_status but it is never reached because the fencing framework decomposes the reboot action into off + poll + on before calling set_power_status. Implement reboot_cycle_fn and pass it to fence_action so that when method=cycle is configured, the agent calls Nova hard reboot API directly.

### DIFF
--- a/agents/openstack/fence_openstack.py
+++ b/agents/openstack/fence_openstack.py
@@ -112,19 +112,15 @@ def set_power_status(conn, options):
         except Conflict as e:
             fail_usage(e)
         logging.info("Called stop API call for " + server.id)
-    if action == "reboot":
-        logging.info("Rebooting instance " + server.name)
-        try:
-            server.reboot("HARD")
-        except Conflict as e:
-            fail_usage(e)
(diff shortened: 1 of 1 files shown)


agents/openstack/fence_openstack.py
```diff
@@ -121,6 +121,14 @@ def set_power_status(conn, options):
         logging.info("Called reboot hard API call for " + server.id)
 
 
+def reboot_cycle(conn, options):
+    server = conn.servers.get(options["--plug"])
+    logging.info("Hard rebooting instance " + server.name)
+    server.reboot("HARD")
```
oalbrigt: I would implement it with the try/except logic that was used in `set_power_status()` and remove that section from there if it's not being used.

oalbrigt (approved):

oalbrigt:
You have to run `make xml-upload` and attach the updated metadata to avoid CI failing due to the metadata change.

oalbrigt:
Thanks.

PR #658, opened 2026-07-08
fence_vmware_rest: add token-based authentication support

arslan9292:
This introduces `--token` and `--token-script` parameters to allow secure, passwordless authentication with the vCenter REST API. This avoids the need to store static passwords or short-lived session IDs in the cluster configuration.

 - Adds dynamic bypass of core credential validation when a token is used.
 - Adds a `fail_usage` check if neither authentication method is provided.
 - Streamlines `connect()` to cleanly handle both token and credential workflows.


oalbrigt:
LGTM.

oalbrigt:
CI fail not related to these changes.

oalbrigt:
Thanks.

PR #659, opened 2026-07-20
fence_gce: fail immediately with import errors

arslan9292:
Previously, if a dependency like `httplib2` failed to load due to
a package conflict, the script would silently pass and crash further
down the stack. Now, it explicitly catches the error, prints a clear
log message identifying the broken module and exits immediately
making aware of the dependency issue.

AI: Assisted by Claude


oalbrigt:
LGTM. CI only fails covscan due to a reset baseline, so that's not related to the patch.

oalbrigt:
Thanks.

PR #660, opened 2026-07-29
fence_vmware_rest: remove static --token option

arslan9292:
A static token provided via `--token` gets invalidated immediately after the agent's first execution as it destroys its session (HTTP DELETE) with vCenter upon disconnect.

Passwordless authentication must use `--token-script` to fetch a fresh session token dynamically on every invocation.


oalbrigt:
LGTM. CI only fails covscan due to a reset baseline, so that's not related to the patch.

oalbrigt:
Thanks.

